PR #3, opened 2017-03-13
Make get_all_nodes context agnostic

tiagojsag:


app/controllers/nodes_controller.rb
```diff
@@ -1,12 +1,22 @@
 class NodesController < ApplicationController
   def get_all_nodes
+
+    context_id = params[:context_id]
+
+    if (not context_id.nil?)
```
agnessa: This can be said without double negation as `if context_id.present?`. Typically parentheses are only used in conditions when there are multiple logical operators and you need to control precedence.

app/controllers/nodes_controller.rb
```diff
@@ -1,12 +1,22 @@
...
     flow_nodes = Flow.select('distinct(unnest(path)) as node_id')
-                     .where('flows.context_id = :context_id', context_id: @context.id)
 
     node_type_ids = NodeType
                         .joins(:context_nodes)
                         .select('node_types.node_type_id')
-                        .where('context_nodes.context_id = :context_id', context_id: @context.id)
+
+    if (context)
+      flow_nodes.where('flows.context_id = :context_id', context_id: context.id)
```
agnessa: You must assign here and in the line below i.e. `flow_nodes = flow_nodes.where...`

tiagojsag:
Fixed and fixed, thanks @agnessa

PR #4, opened 2017-03-15
Change Layer and LayerGroup to ContextLayer and ContextLayerGroup; fi…

tiagojsag:
…lter context layer group loading per context


app/controllers/map_controller.rb
```diff
@@ -1,16 +1,17 @@
 class MapController < ApplicationController
 
   def index
-    dimension_groups = LayerGroup.select('distinct name, layer_group.id as id, layer_group.position')
-                       .joins('LEFT JOIN layer ON layer.layer_group_id = layer_group.id')
-                       .order('layer_group.position ASC, layer_group.id ASC')
+    dimension_groups = ContextLayerGroup.select('distinct name, context_layer_group.id as id, context_layer_group.position')
+                           .joins('LEFT JOIN context_layer ON context_layer.context_layer_group_id = context_layer_group.id')
+                           .where('context_layer_group.context_id = :context_id', context_id: @context.id)
```
tiagojsag: this should be the only functional change

PR #6, opened 2017-03-16
remove support for country and commodity params

tiagojsag:


app/controllers/application_controller.rb
```diff
@@ -11,38 +11,15 @@ class ApplicationController < ActionController::Base
...
-    country_name = params[:country]
-    country_id = params[:country_id]
-    commodity_name = params[:commodity]
-    commodity_id = params[:commodity_id]
 
-    if (not context_id.nil?)
+    if context_id.present?
       @context = Context.find(context_id)
 
       raise ActionController::ParameterMissing, 'The provided context could not be found' if @context.nil?
```
agnessa: actually this line will never execute - if context is not found, find will throw an exception

spec/controllers/download_controller_spec.rb
```diff
@@ -3,17 +3,20 @@
 RSpec.describe DownloadController, type: :controller do
   include_context "two flows"
   describe "GET index" do
+    before(:all) do
+      context = Context.first
```
agnessa: this is probably not necessary (same in the other controller spec) - context variable is defined in the included rspec context and so in this situational context (!) probably no need to redefine it ;) if it ends up being nil maybe there's some other problem

app/controllers/application_controller.rb
```diff
@@ -11,38 +11,15 @@ class ApplicationController < ActionController::Base
...
-      raise ActionController::ParameterMissing, 'Required country missing' if country_name.nil? and country_id.nil?
-      raise ActionController::ParameterMissing, 'Required commodity missing' if commodity_name.nil? and commodity_id.nil?
-
-      if (not country_id.nil? and not commodity_id.nil?)
-        @context = Context
-                       .where('commodity_id = ?', commodity_id.to_i)
-                       .where('country_id = ?', country_id.to_i)
-                       .first
-      else
-        @context = Context
```
agnessa: love to see it gone, really hated this `ILIKE` bit :)

tiagojsag:
Done, thanks for the feedback

PR #18, opened 2017-05-18
Puts checks in place where values might be missing

### LINKED ISSUE #17: Error 500 on place factsheet endpoint
http://staging.trase.earth/api/v2/get_place_node_attributes?&node_id=15606

agnessa:
Resolves #17 - apparently for some municipalities the biome value might be missing (applied the same check to state)


agnessa:
@tiagojsag looks good to merge?

tiagojsag:
looks like it, yes

PR #22, opened 2017-05-26
remove unused node hierarchy

tiagojsag:


agnessa:
Please remove the definition of `ancestors` from Node and amend lines 26-29 in app/models/place_attributes.rb:

```
    ([@node] + @node.ancestors.to_a).each do |node|
      @data[(node.node_type.node_type.downcase + '_name').to_sym] = node.name
      @data[(node.node_type.node_type.downcase + '_geo_id').to_sym] = node.geo_id
    end
```
with sth like:

```
      @data[(@node.node_type.node_type.downcase + '_name').to_sym] = @node.name
      @data[(@node.node_type.node_type.downcase + '_geo_id').to_sym] = @node.geo_id
```

PR #26, opened 2017-06-13
Do not filter non-temporal indicator values by year for ranking

agnessa:
I believe it helps with most indicators (e.g. water scarcity, fires), still need to check on some odd empty rows in "other environmental indicators" but this at least should sort out the default selected tab.

https://basecamp.com/1756858/projects/12498794/todos/308418561


agnessa:
Suppressed the empty rows.

PR #34, opened 2017-09-07
Update profile page logic for new data

tiagojsag:
https://github.com/sei-international/TRASE/blob/new-data-2017/DB/mapping_variables

some quants/inds were renamed, others moved around, so it's api changing time!

### DIFF
--- a/app/models/ind.rb
+++ b/app/models/ind.rb
@@ -30,6 +30,6 @@ class Ind < ActiveRecord::Base
     has_many :context_indicators, as: :indicator
 
     scope :place_temporal, -> {
-      where("place_factsheet AND place_factsheet_temporal OR name IN ('POTENTIAL_SOY_RELATED_DEFOR_ind', 'TOTAL_DEFOR_RATE')")
+      where('place_factsheet AND place_factsheet_temporal')
     }
 end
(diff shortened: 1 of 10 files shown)


agnessa:
🤦‍♀️

app/models/ind.rb
```diff
@@ -30,6 +30,6 @@ class Ind < ActiveRecord::Base
     has_many :context_indicators, as: :indicator
 
     scope :place_temporal, -> {
-      where("place_factsheet AND place_factsheet_temporal OR name IN ('POTENTIAL_SOY_RELATED_DEFOR_ind', 'TOTAL_DEFOR_RATE')")
+      where('place_factsheet AND place_factsheet_temporal')
```
agnessa: Are you sure `POTENTIAL_SOY_RELATED_DEFOR_ind` is no longer required here? This is clearly not a part of code I've written that I'm particularly proud of, but while I understand that `TOTAL_DEFOR_RATE` not only changed names but also changed from `ind` to `qual`, whereas `POTENTIAL_SOY_RELATED_DEFOR_ind` is still an ind and maybe relevant here? :D

tiagojsag: POTENTIAL_SOY_RELATED_DEFOR_ind is gone from the new db entirely

agnessa: oh in that case maybe it needs removing also from `place_attributes.rb`?

tiagojsag: seems I missed that and a couple more things :s

PR #40, opened 2017-10-02
ACTOR: Calculation of sustainability indicator totals for top sources

agnessa:
Please merge after #39 - which implements the `node#flow_values` method.

This restores calculation of sustainability indicators from flow rather than node values.
https://basecamp.com/1756858/projects/12498794/todos/325801068


tiagojsag:
LGTM, pending confirmation on https://basecamp.com/1756858/projects/12498794/todos/326171705

PR #43, opened 2017-10-03
Actor profile - Replace exporter data in importer page; Use dynamic r…

tiagojsag:
…ounding in volume produced

### DIFF
--- a/app/models/actor_attributes.rb
+++ b/app/models/actor_attributes.rb
@@ -59,9 +59,26 @@ class ActorAttributes
   def initialize_trade_volume_for_summary
     trade_flows_current_year = @node.flow_values(@context, 'quant', 'Volume').where(year: @year)
     @trade_total_current_year = trade_flows_current_year.sum('value')
+    case
+      when @trade_total_current_year < 1000
+        trade_total_current_year_value = @trade_total_current_year
+        trade_total_current_year_unit = 'tons'
+        trade_total_current_year_precision = 0
+      when @trade_total_current_year < 1000000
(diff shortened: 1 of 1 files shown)


app/models/actor_attributes.rb
```diff
@@ -209,8 +226,17 @@ def sustainability
   end
 
   def companies_exporting
+    quant = Quant.find_by(:name => 'SOY_')
+
+    unit = quant.unit
+    value_divisor = 1
+    if (unit.casecmp('tn') == 0)
```
tiagojsag: This feels like a less hacky hack than simply dividing by 1k all the time.

tiagojsag:
requires https://github.com/Vizzuality/TRASE-frontend/pull/474

app/models/actor_attributes.rb
```diff
@@ -270,13 +296,13 @@ def companies_exporting
...
         id: total['node_id'],
-        y: total['value'].to_f / 1000, # TODO hack
+        y: total['value'].to_f / value_divisor,
         x: indicator_totals_hash[total['node_id']]
       }
     end
 
     {
-      companies_exporting: {
+      companies_sourcing: {
```
agnessa: not sure why we needed to rename this?

app/models/actor_attributes.rb
agnessa: yes

app/models/actor_attributes.rb
tiagojsag: we dont need, but feels itchy

PR #48, opened 2017-10-03
Removed SOY_ quant from exporting companies scatterplot

agnessa:
Removes references to `SOY_` quant in remaining profile calculations.
- replaces `node_quants` with `flow_quants` calculations in the "exporting companies" scatterplot
- removes a top-level attribute `total_soy_2015` which seems obsolete

https://basecamp.com/1756858/projects/12498794/todos/325807711


agnessa:
@nerik please confirm it is OK to remove the `total_soy_2015` field from the API  - see https://github.com/Vizzuality/TRASE-frontend/issues/475

PR #51, opened 2017-10-05
Actor profile - Load units and names of sustainability indicators fro…

tiagojsag:
…m database

### DIFF
--- a/app/models/actor_attributes.rb
+++ b/app/models/actor_attributes.rb
@@ -364,9 +364,9 @@ is #{@main_destination_name.humanize}, accounting for \
 
   def risk_indicators
     [
-      {name: 'Territorial deforestation', backend_name: 'DEFORESTATION_V2'},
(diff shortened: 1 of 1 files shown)


app/models/actor_attributes.rb
```diff
@@ -417,7 +417,12 @@ def sustainability_for_group(name, node_type, include_totals)
     end
     {
       name: name,
-      included_columns: [{name: node_type.humanize}] + risk_indicators.map{ |indicator| indicator.slice(:name, :unit) },
+      included_columns:
+          [{name: node_type.humanize}] +
+          risk_indicators.map {
+              |indicator| quant = Quant.find_by(name: indicator[:backend_name])
+            {name: indicator.key?(:name) ? indicator[:name] : quant.frontend_name , unit: quant.unit}
```
tiagojsag: AGROSATELITE_SOY_DEFOR_ has a slightly different label in the database, so for now we need to keep this fallback logic here. Maybe this is something to address on the new db structure

PR #53, opened 2017-10-06
Add version per context

tiagojsag:


db/migrate/20171006171936_add_data_download_version_per_context.rb
```diff
@@ -0,0 +1,5 @@
+class AddDataDownloadVersionPerContext < ActiveRecord::Migration[5.0]
+  def change
+    add_column :download_versions, :context_id, :integer
```
agnessa: in addition to adding the column we could already add rows for all contexts as part of migration and set the column as `NOT NULL`

app/models/flow_download.rb
```diff
@@ -13,7 +13,7 @@ def initialize(context, params)
     else
       ','
     end
-    @download_name = "#{@context.country.name}_#{@context.commodity.name}_#{DownloadVersion.current_version_symbol}"
+    @download_name = [@context.country.name, @context.commodity.name, DownloadVersion.current_version_symbol(@context)].reject(&:nil?).join('_')
```
agnessa: `reject(&:nil?)` can be expressed more shortly as `compact`

db/migrate/20171006171936_add_data_download_version_per_context.rb
tiagojsag: They requested that we also support contexts with no version, so we either make version optional at a database by having no row, or we need to allow null values. I believe no row is easier to maintain, otherwise it will require an explicit "empty" row, which is weird.

agnessa: I'd go with no row. Do we have the list of which versions / no versions to use and where?

tiagojsag: that's what we have. I also changed the logic that loads version to account for this.

current config here: https://basecamp.com/1756858/projects/12498794/todos/326870870

app/models/flow_download.rb
tiagojsag: fixed

agnessa:
👍

PR #56, opened 2017-10-14
Fix schema migration script + Response schema testing + minor changes in responses

tiagojsag:


tiagojsag:
I created a dump that works with the migration script here: https://github.com/sei-international/TRASE/tree/revamped_schema

It removes duplicated values on fires (which had different values, so I removed the higher ones) and flow_quals (which were plainly duplicated)

tiagojsag:
Changes in get_linked_geo_ids response require https://github.com/Vizzuality/TRASE-frontend/compare/change-get_linked_geo_ids-response-handling-to-account-for-new-response-format?expand=1 in the FE

agnessa:
I would really like to merge the first commit with the script fix - is this PR still in WIP? in that case any chance of splitting the PR?

tiagojsag:
I won't have time to work on it anymore, but each commit is atomic, we just have to be careful with the last one, as it requires FE changes

agnessa:
I merged the script fix commit into the revamp branch, and I'll merge the rest into develop but must open a new PR with a rebased version.

agnessa:
https://github.com/Vizzuality/trase-api/pull/58

PR #61, opened 2017-10-21
Replace get_nodes with get_node_attributes

tiagojsag:
Requires https://github.com/Vizzuality/TRASE-frontend/pull/493

Old API performance for year 2015 context_id 1:
8.75MB
<12s response time

Brand new implementation awesome values:
5.77MB
<6s response time

This is not a drop-in replacement for the old get_nodes, but loads the same data, ignoring some unnecessary values.

This is a PR into my other tests branch, but can most likely be cherry-picked into develop directly, except for the json schema validation part.


tiagojsag:
Added https://github.com/ohler55/oj which helps cut down the json encoding time by 1s

tiagojsag:
Added a "validate:get_node_attributes" rake task that compares the results with some reference files I generated from the old API. It needs the full db, and the same version it was used to generate the reference files, so this is why I didn't add this to travis.

app/controllers/nodes_controller.rb
```diff
@@ -26,4 +26,29 @@ def get_all_nodes
...
+      raise ActionController::ParameterMissing, 'Required start_year missing'
+    end
+
+    unless params[:end_year].present?
+      raise ActionController::ParameterMissing, 'Required end_year missing'
+    end
+
+    node_ids = Node
+                   .select('node_id')
+                   .joins('LEFT JOIN context_nodes ON nodes.node_type_id = context_nodes.node_type_id')
```
agnessa: this could be just inner join I think

app/controllers/nodes_controller.rb
```diff
@@ -26,4 +26,29 @@ def get_all_nodes
...
+                   .where('context_nodes.context_id = ?', @context.id)
+
+    ind_values = NodeInd.get_attributes_for_nodes(node_ids, @context.id, start_year, end_year)
```
agnessa: I don't like this approach of first retrieving the node ids and then putting them in an enormous IN clause - such queries tend to be slow. Could we try what happens if we changed get_attributes_for_nodes into get_attributes_for_context, so that the queries in NodeInd and NodeQuant perform an inner join with context_nodes instead?

lib/tasks/validate_get_node_attributes.rake
```diff
@@ -0,0 +1,55 @@
...
+      unless (things.first['rawValue'].present? and api_value.value.present? and things.first['value3'].present? and api_value.bucket3.present? and things.first['value5'].present? and api_value.bucket5.present?)
+        raise "Missing data for #{api_value.node_id}. Cannot compare"
+      end
+
+      if (things.first['rawValue'] != api_value.value or things.first['value3'] != api_value.bucket3 or things.first['value5'] != api_value.bucket5)
+        raise "Node #{api_value.node_id} has mismatching values for attribute type #{api_value.attribute_type} with attribute id #{api_value.attribute_id}. This most likely means a bug in get_node_attributes"
+      end
+    end
+
+    puts 'Congratulations, your code is bug-free.'
```
agnessa: I guess it will take us quite a while to write those tests against all reference responses in this way. I thought it would be much faster to cover all endpoints by using a json diffing tool here. What would be the drawbacks?

app/controllers/nodes_controller.rb
tiagojsag: probably, will test

app/controllers/nodes_controller.rb
tiagojsag: :+1:

tiagojsag:
As per my conversation with @agnessa we're freezing this PR for a while, until we are confident we can handle another moving part in the project

agnessa:
@tiagojsag will you have time today to look at this? otherwise I'm happy to do it in order to unblock an upcoming task.

tiagojsag:
I'll try, let's chat about it

2 things:
- There's a rake task that transforms one of the responses into the other and validates that they match. This was failing in my latest iteration, didn't have time to look into it
- There's a bug in the old API implementation of this function, which lead it to return less data than expected. So our enpoint gets extra points for awesomeness

tiagojsag:
@agnessa ready for review

db/structure.sql
```diff
@@ -14,55 +7,6 @@ SET check_function_bodies = false;
 SET client_min_messages = warning;
 SET row_security = off;
 
---
--- Name: revamp; Type: SCHEMA; Schema: -; Owner: -
---
-
-CREATE SCHEMA revamp;
```
agnessa: All of `revamp` schema gets removed when you run `rake db:migrate`, sadly because of the search path set to `public` on the default database connection. To avoid issues with messing up the schema temporarily comment out the search path setting in `database.yml` when running migrations.

agnessa:
build is failing because of messed up `structure.sql`, please restore

PR #62, opened 2017-10-24
Api versioning

tiagojsag:


app/controllers/api/v3/nodes_controller.rb
```diff
@@ -0,0 +1,32 @@
+module Api
```
agnessa: Can we move this to a different PR? There are too many changes already in this one and it still requires more work to make it mergable, so let's not add to it anymore.

tiagojsag:
@agnessa ready for review and merge

app/controllers/api/v3/v3_application_controller.rb
```diff
@@ -0,0 +1,19 @@
+module Api
+  module V3
+    class V3ApplicationController < ApplicationController
```
agnessa: is this V3 prefix needed in the class name?

app/serializers/api/v3/get_contexts/recolor_by_attribute_serializer.rb
```diff
@@ -0,0 +1,26 @@
...
+        attribute :tooltip_text, key: :description
+
+        attribute :type do
+          object.original_type.downcase
+        end
+
+        attribute :nodes do
+          flows = []
+          if (object.legend_type.eql? 'qual')
+            FlowQual.distinct.select('value').where('qual_id = ? AND value NOT ILIKE \'UNKNOWN%\'', object.original_id).each do |flow|
```
agnessa: this looks dodgy, as in this whole pushing in the loop could be replaced by a pluck no? the naming is also quite confusing to me; what this loop produces is an array of flow_qual values and it's in a variable called `flows`, returned to the client as `nodes`; at this point we don't want to change the response, but it would be great to document what this does and maybe we could revisit it when we're done with stabilising.

config/database.yml
```diff
@@ -12,6 +6,12 @@ default: &default
   port: <%= ENV.fetch("POSTGRES_PORT") { "5432" } %>
   password: <%= ENV.fetch("POSTGRES_PASSWORD") {} %>
   database: <%= ENV.fetch("POSTGRES_DATABASE") { "trase_dev" } %>
+  schema_search_path: "public"
+
+revamp:
```
agnessa: we don't need this anymore right?

app/controllers/api/v3/v3_application_controller.rb
tiagojsag: Most likely it's not, I'll remove it

app/serializers/api/v3/get_contexts/recolor_by_attribute_serializer.rb
tiagojsag: dodgy is my code's middle name ;) Just to mention that this is not new code, although git sees it that way. But I'm sure there are 1001 improvement points all over the place, this is some of my very first rails code ;)

config/database.yml
tiagojsag: will remove

app/serializers/api/v3/get_contexts/recolor_by_attribute_serializer.rb
agnessa: Sure, I appreciate that is the case :) I think when there are small readability improvements to be made we should be making them as we migrate the code, and when there are bigger parts of hairy code we could flag them with # TODO: refactor or something to make sure we don't forget about them when we have stabilised the thing. This is a great opportunity to review our whole codebase, even if some things will need to wait to be fixed.

tiagojsag: We actually don't need this here, as this is controller-related logic. moved to the other PR

tiagojsag:
Removed some unnecessary code, rebased, the usual. Should be good to merge

agnessa:
Can we remove the V2 base controller as well?

PR #64, opened 2017-10-31
Added gold master tests for csv downloads and json responses

agnessa:
Adds a rake task to run gold master tests, which validate system outputs against known valid versions of those outputs for same input parameters.

Those tests are intended to be run against a full database.

The file `spec/support/gold_master_urls.yml` contains a list of urls that generate outputs. Those are relative urls and they include csv and json outputs of the system.

There are 2 rake tasks:

- `bundle exec rake gold_master:record` - this one should be re-run when `spec/support/gold_master_urls.yml` is updated. It records the gold master responses for all the urls and zips them in `spec/support/gold_master.zip`. It should be run using the pre-revamp version of the backend code and database:
    * https://github.com/sei-international/TRASE/releases/tag/pre-revamp
    * https://github.com/Vizzuality/trase-api/releases/tag/pre-revamp
    * Note: some responses are huge, over 1 MB. They're zipped and stored using GLFS.
- `bundle exec rake gold_master:test` - this one collects responses as generated by the current version of the code and compares with the gold master using json & csv diffing tools. It's intended to be used with the same version of the database as the gold master.
  * Note: the responses are stored uncompressed in `tmp/actual` and are not cleaned after the tests have run.

Both tasks are parametrised by same env variables to specify the hosts on which to run, when running in local environment this is going to be something like:

```
GOLD_MASTER_HOST_V1=http://localhost:8080
GOLD_MASTER_HOST_V2=http://localhost:3000
GOLD_MASTER_HOST_V3=http://localhost:3000
```


tiagojsag:
Let's discuss this personally or on a call, I have some questions about the workflow you have in mind for this.

agnessa:
Added sorting of hashes before comparison

tiagojsag:
LGTM as soon as travis approves

PR #65, opened 2017-10-31
Added rubocop for sanity

agnessa:
Adds [rubocop](https://github.com/bbatsov/rubocop) with slightly modified list of rules. I think we might want to refine the rules as we go.

There are currently a large number of style violations in the code - those are listed in `.rubocop_todo.yml`. Many of them will not be relevant anymore as we migrate the code. The important thing is not to migrate the style violations into the new code, which is where this will help.

It's possible to integrate Hound CI to get code style violation notifications e.g. on submitting a pull request. That generates a lot of noise, which I'm fine with because it's obviously a good motivation to run `rubocop` and fix violations before pushing. So I would be keen to enable that, and also we could probably take advantage of a pre commit hook that could run `rubocop -a` - autocorrections. Happy to discuss how to best do it, this PR just adds rubocop + rules/exceptions.


nerik:
Tiago will have more to say on this, but it should be doable to apply the linter only commits' delta, no?

agnessa:
Yes, this is definitely how Hound CI works.

agnessa:
`rubocop` itself can be instructed to exclude files with known violations from particular checks - this is the purpose of the `rubocop_todo.yml` file. When I run rubocop locally it will currently report no violations because all the old ones from that file are excluded. This way it can be used to only monitor for violations in the newly modified code.

tiagojsag:
Just tested, seems great :)
For the time being, I'm happy with just doing a full check - the main reasoning behind the partial check we do on other projects was that a full one was too slow.

tiagojsag:
I just let the dog out (omg such bad pun) and it's probably spamming away on our (my aweful) PRs as I type.

I'm wondering if it doesn't make sense to use travis instead - achieves the same goal, but keeps the PR discussions decluttered

tiagojsag:
we might want to cherry-pick https://github.com/Vizzuality/trase-api/pull/67/commits/1e3b7ea7427b0a63d75293b9ba55f6b9e695ab86

agnessa:
rebased and resolved conflicts, cherry picked Hound CI configuration

tiagojsag:
thoughts on HoundCI vs Travis for CS validation?

agnessa:
I'm happy with both, maybe a bit more on the hound side because the travis build takes a while, whereas hound is very quick to complain as soon as you create a PR, so you can start fixing right away.

PR #69, opened 2017-11-03
Versioned controllers response comparison

tiagojsag:


agnessa:
I think this is now OK to merge. This only adds one endpoint, `get_contexts`, but with both unit tests and gold master tests passing.

PR #70, opened 2017-11-03
Sending test coverage to Code Climate

agnessa:
testing PR with coverage report


tiagojsag:
https://codeclimate.com/github/codeclimate/codeclimate/badges

agnessa:
So my understanding is if we merge this, we'll have a coverage metric on our code climate dashboard, which is now missing.

screen shot 2017-11-06 at 10 41 52

tiagojsag:
Perhaps CC needs a but more configuring. And while it's not a show stopper, I also see a lot of weird analysis from CC on other PRs, which I am not sure are useful ATM. Is there a way to disable those, until we can commit ourselves to fixing them?

agnessa:
Yes, analysis can be configured using the config file I just added - which currently turns on some extra analysis plugins. Here's a description of what can be fine tuned and how:
https://docs.codeclimate.com/v1.0/docs/configuring-your-analysis

PR #71, opened 2017-11-09
Integrate frontend code into this project

tiagojsag:


tiagojsag:
For the time being, I think I want to keep the actual integration to a minimum - right now, the two code bases only share a readme and the git hooks
As we move forward, we might want to add closer integration - I see a unified deploy process as something to work on soon - but I think we should tackle that separately.

Thoughts?

agnessa:
I'm fine with that

nerik:
LGTM 👍

PR #72, opened 2017-11-09
Feature/make rubocop happy again

tiagojsag:


app/models/flow_download_query_builder.rb
```diff
@@ -5,17 +5,18 @@ def initialize(context, params)
     initialize_path_column_names(@context.id)
     @query = @query.where(year: params[:years]) if params[:years].present?
     if params[:indicators].present?
-      tmp = [Quant, Ind, Qual].map do |indicator_class|
+      # TODO: I highly suspect that this .map block is useless
```
tiagojsag: @agnessa need your input here please

agnessa: yes

PR #75, opened 2017-11-17
Feature/schema revamp continued

agnessa:
Added new "properties" tables in order to remove "yellow" columns from "blue" data tables (an annoying colour metaphor which I'm afraid I'll keep using):
- country_properties
- context_properties
- context_node_type_properties
- node_properties
- ind_properties
- qual_properties
- quant_properties

Some other cleanup was done related to recent discussions:
- traders table removed, replaced by main_id in nodes
- parent_id removed from commodities, as we're not implementing that yet and probably this wouldn't work anyway
- removed updated_at timestamps from "blue" tables, as they are never updated


agnessa:
I'll figure out how to override reek's complaints and the reported coverage issues another time :)

tiagojsag:
too much to wrap my head around in one go :S guess I'll merge, and we can improve as we go forward

PR #77, opened 2017-11-20
Staging deploy

tiagojsag:
- bump ruby version (not really necessary, but hey, it helps)
- run npm install, cp env file and run npm run build on capistrano deploy
- a few more things that are needed to keep capistrano happy


agnessa:
I tried running the specs locally and the build fails on ruby 2.4.2, passes on 2.4.1. I think you haven't changed the version of ruby for travis, which is why CI still passes.

tiagojsag:
Yeah, I had fixed it, just on another branch. Now it's here, and I also updated travis

agnessa:
great, thanks!

config/deploy.rb
```diff
@@ -32,9 +32,27 @@
...
-set :rvm_ruby_version, '2.4.1'
\ No newline at end of file
+namespace :npm do
+  after 'npm:install', 'npm:build'
+
+  task :build do
+    on roles fetch(:npm_roles) do
+      within fetch(:npm_target_path, release_path) do
+        with fetch(:npm_env_variables, {}) do
+          upload! "frontend/.env.#{fetch(:stage)}", "#{current_path}/frontend/.env"
```
agnessa: does this upload a local `.env` file? could we not do it following the same principle as the backend .env file, which is kept on it's target environment between releases?

config/deploy.rb
```diff
@@ -32,9 +32,27 @@
...
+
 namespace :deploy do
   after :finishing, 'deploy:cleanup'
   after 'deploy:publishing', 'deploy:restart'
 end
 
-set :rvm_ruby_version, '2.4.1'
\ No newline at end of file
+namespace :npm do
+  after 'npm:install', 'npm:build'
```
agnessa: what is the full deployment command now?

config/deploy.rb
tiagojsag: :+1:

config/deploy.rb
tiagojsag: same as before, cap <env> deploy.

tiagojsag:
@agnessa ready for review

PR #78, opened 2017-11-20
Change frontend to support new content response format

tiagojsag:
FE Changes:
- Made FE content calls depend on the API endpoint settings, which are now unified. This means less API endpoints to manage, and a more centralized way to handle these sort of thing
- Minor changes to some responses (casing of attributes and a "data" element at the root of the response)

BE Changes:
- Use a serializer on tweets


frontend/scripts/utils/getURLFromParams.js
```diff
@@ -45,6 +51,19 @@ function getURLForV2(endpoint, params = {}) {
   }, `${API_V2_URL}${endpoint}?`);
 }
 
+function getURLForContent(endpoint, params = {}) {
```
nerik: This file needs to be refactored as there is now 4 times more or less the same logic to build endpoints URLs.

tiagojsag: done. Pretty soon we'll have a single API, so we'll simplify it even more

PR #83, opened 2017-11-21
fixed v2 endpoint and refactored URL building logic

nerik:
Fixes api/v2/api/v2 calls.
Additionally, DRYed getURLFromParams.js

It seems that content APIs have been moved from `/content` to `/api/v2/content` but not deployed yet, right?


tiagojsag:
With API v3, this file will most likely become a lot simpler, so I think it's best if we keep it as is, to keep different API logic separated and easy to phase out.

PR #84, opened 2017-11-22
Remove basic auth from FE webserver

tiagojsag:


nerik:
@tiagojsag Is there a particular reason why we keep a dev server in express ? (in `/frontend/config/application.js`)
As far as I know it's not in use as we are using a webpack server for dev.

tiagojsag:
I think I did a major copy-paste fro GFishW at some point, so we could host the FE on heroku. Then we moved on to our own servers, and it was just left there. I'll clean up some more

nerik:
Just did it, also changed `npm run dev` to a more standard `npm start`

tiagojsag:
@nerik thanks

I also cleaned up the package.json a bit. We had too many actions, many of which were duplicates or legacy. Let me know your thoughts, and merge if it LGMT

frontend/package.json
```diff
@@ -2,18 +2,11 @@
   "name": "trase-frontend",
   "version": "1.0.0",
   "description": "",
-  "main": "server.js",
   "scripts": {
-    "test": "echo \"Error: no test specified\" && exit 1",
-    "start": "npm run webpack:server",
-    "dev": "npm run start",
```
nerik: I was keeping this one as an alias for a "soon to be deprecated" command, but up to you.

frontend/package.json
```diff
@@ -2,18 +2,11 @@
...
-    "build": "npm run webpack:prod && cp -a public/. dist/",
-    "webpack:prod": "NODE_ENV='production' webpack --config config/webpack.js -p",
-    "webpack:server": "webpack-dev-server --inline --port 8081 --host 0.0.0.0 --content-base public/ --config config/webpack.js",
-    "eslint": "npm run lint-js",
+    "start": "webpack-dev-server --inline --port 8081 --host 0.0.0.0 --content-base public/ --config config/webpack.js",
+    "build": "NODE_ENV='production' webpack --config config/webpack.js -p && cp -a public/. dist/",
```
nerik: I prefer having separated NPM tasks to have a more readable package.json. Agglomerating tasks into single lines tend to make them long and difficult. I'd keep `"webpack:prod"`

frontend/package.json
```diff
@@ -2,18 +2,11 @@
...
   "version": "1.0.0",
   "description": "",
-  "main": "server.js",
   "scripts": {
-    "test": "echo \"Error: no test specified\" && exit 1",
-    "start": "npm run webpack:server",
-    "dev": "npm run start",
-    "build": "npm run webpack:prod && cp -a public/. dist/",
-    "webpack:prod": "NODE_ENV='production' webpack --config config/webpack.js -p",
-    "webpack:server": "webpack-dev-server --inline --port 8081 --host 0.0.0.0 --content-base public/ --config config/webpack.js",
```
nerik: Same here, I would keep `"webpack:server` and `start` separated. This way it's easy at a first glance to identify what `npm start` does.

nerik:
See comments above. In a nutshell: remove unneeded aliases, why not, agglomerate into long one-liners, I'd rather not.

PR #86, opened 2017-11-22
New search

nerik:
The goal of this PR is to replace the current vanilla JS/awesomeplete search component with a react/Downshift based solution, and to add the new search features in the designs (the latter being made possible by the former).


nerik:
Letting @sorodrigo pick this up to allow him to get into some core TRASE functionality.

TODO:
- [ ] fix the open/close mechanism of the search
- [ ] implement external links ("see importer profile", etc). For this you'll need to create a version of the node list where importer and exporters are merged to a single node, and create buttons depending on the node type (2 buttons for importer/exporter, 1 for municipalities, 0 for the rest)
- [ ] implement "add to supply chain" button. While the default functionality of the autocomplete line is to yield a new sankey with a single node selected, this button will create an additive selection, similar to what happens when clicking on several polygons on the map
- [ ] implement a compact version of the "nodes selected bar" on the bottom of the sankey, to appear before the text input
- [ ] make a version of the search component that is flexible enough to be used on the profile search page, and use it there. Don't forget to cleanup the old search component when done (JS, CSS, ejs template).
- [ ] profit.

frontend/scripts/actions/tool.actions.js
```diff
@@ -463,15 +463,13 @@ function getSelectedNodeIds(currentSelectedNodesIds, changedNodeId) {
 
 export function selectNode(nodeId, isAggregated = false) {
   return (dispatch, getState) => {
+    const { tool } = getState();
     if (isAggregated) {
       dispatch(selectView(true));
+    } else if (tool.areNodesExpanded) {
+      dispatch(selectExpandedNode(nodeId));
     } else {
-      const currentSelectedNodesIds = getState().tool.selectedNodesIds;
```
tiagojsag: Just tested this, this is a behavior change, and I'm not sure we have discussed it
- Start with a clean sankey
- Select "Bungee" and "Cargil" from exporter column
- Expand by those 2 nodes
- Select "Bungee" from importer column

In the previous code, this filters the flows without recomputing the sankey
With these changes, this recomputes the sankey

image

frontend/scripts/constants.js
```diff
@@ -1,3 +1,6 @@
+export const ACTORS_COLUMN_IDS = [6, 7];
```
tiagojsag: evil evil evil!
you are hardcoding things that can easily change, effectively tying the frontend to a specific database value.

frontend/scripts/containers/shared/search-react.container.js
```diff
@@ -0,0 +1,49 @@
...
+  const { nodes, selectedNodesIds } = state.tool;
+  // store nodes at container level to avoid rerendering when filtering... for want of a better solution
+  if (nodes !== undefined && (!searchNodes || nodes.length !== searchNodes.length)) {
+    const allNodes = nodes.filter(
+      node => node.hasFlows === true &&
+      node.isAggregated !== true &&
+      node.isUnknown !== true
+    );
+
+    searchNodes = Object.values(groupBy(allNodes.filter(x => ACTORS_COLUMN_IDS.includes(x.columnId)), 'mainNodeId'))
```
tiagojsag: allNodes already specifies which type of profile each node has, so you can use that instead of the 3xEvil constant

frontend/scripts/containers/shared/search-react.container.js
```diff
@@ -0,0 +1,49 @@
...
+      .map(([nA, nB]) => nB ?
+        ({
+          id: `${nA.id}_${nB.id}`,
+          name: nA.name,
+          type: IMPORTER_EXPORTER_TYPE,
```
tiagojsag: We can also remove this (also evil) constant if we group nodes by main_node_id instead of looking for specific patterns based on hardcoded, unflexible constants

frontend/scripts/react-components/shared/search.component.js
```diff
@@ -0,0 +1,221 @@
...
+import { findAll } from 'highlight-words-core';
+import cx from 'classnames';
+
+import { PROFILE_PAGES_WHITELIST } from 'constants';
+
+import NodeTitleGroup from 'containers/shared/node-title-group-react.container';
+
+export default class Search extends Component {
+
+  static getNodeId(selectedItem) {
```
tiagojsag: didn't even know JS had static methods. Anyway, is there any reason for this particular method to be static? Looks like it should be "private", if anything

frontend/scripts/react-components/shared/search.component.js
```diff
@@ -0,0 +1,221 @@
...
+    // TODO: implement dual selection, for now just displays importer
+    const parts = (selectedItem.id + '').split('_');
+    return parts.length > 1 ? parseInt(parts[0]) : selectedItem.id;
+  }
+
+  static isValidChar(key) {
+    return (/^([a-z]|[A-Z]|ñ|Ñ|á|é|í|ó|ú|Á|É|Í|Ó|Ú){1}$/.test(key));
```
tiagojsag: my keyboard has ß, ö ä an ü, and I just broke your regex ;)
Portuguese alphabet also has ç ã õ and a few others which will also break your regex - and municipality names have all of those.
Rule of thumb: never use your own regexes to do this sort of thing. I'm sure that there's a npm package out there we can use to do this for us.

frontend/scripts/react-components/shared/search.component.js
```diff
@@ -0,0 +1,221 @@
...
+                          <div class='node-actions-container'>
+                            <button
+                              onClick={e => this.onAddNode(e, item)}
+                              class='c-button -medium-large'
+                              disabled={itemSelected}
+                            >
+                              {itemSelected ? 'Already in' : 'Add to'} supply chain
+                            </button>
+                            {
+                              PROFILE_PAGES_WHITELIST.includes(item.type) && item.type.split(' & ')
```
tiagojsag: this is isdented wwwwaaayyyy to deep, which is a smell. I would either:
- split the render method into smaller methods 
- pre-compute all these "map" and "filter" results before. 

We should aim at having JS decoupled from HTML as much as possible, to make things easier to read

frontend/scripts/utils/getURLFromParams.js
```diff
@@ -95,7 +95,7 @@ export function getURLFromParams(endpointKey, params = {}, mock = false) {
       case 3:
         return getURLForV3(`/api/v3${endpointData.endpoint}`, params);
       case 2:
-        return getURLForV2(`/api/v2${endpointData.endpoint}`, params);
```
tiagojsag: nope nope nope nope

frontend/.babelrc
```diff
@@ -1,6 +1,7 @@
 {
   "presets": [ "es2015" ],
   "plugins": [
+      "transform-object-rest-spread",
```
tiagojsag: I'm not saying no to this particular plugin, but let's try to keep them to a minimum. They do offer awesome things, yes, but let's also keep in mind that they simply transpile code into native JS, with varying levels of performance

sorodrigo: I agree that we should consider before using plugins that end up using heavy polyfills. But for all others that just end up transpiling to native code I disagree. The speed and complexity it sometimes removes is greater than the performance loss, to me it smells like premature optimization.

tiagojsag: very true, we need to keep a good balance of both, taking into account all pros and cons :+1:

frontend/scripts/actions/tool.actions.js
sorodrigo: Indeed, this is a change in behaviour. The thing is, that having the ability to select a node that isn't visible through search exposed and existing issue (previously limited to the map selection). Specifically what happened was the following:
1. Selected through the map a non visible municipality node.
2. Select an exporter.
3. Select a country.
If you deselected the original municipality node the result was all flows from that (now deselected municipality to the selected exporter/country). This to me makes no sense, the resulting flow should be all municipalities that pass through the selected exporter and through the selected country.
I showed this to @juancarlosalonso and we decided that expanded view should always recalculate on adding/removal of nodes.

frontend/scripts/constants.js
sorodrigo: Totally agree. Given that this was an ID I supposed it was more fixed that the type prop. Will change to `type` property. 👍

frontend/scripts/containers/shared/search-react.container.js
sorodrigo: The `mainNodeId` is already used for the grouping. This is just a string used for reflecting nodes as grouped in the result list. And to filter the results that have a link to profile pages.

frontend/scripts/react-components/shared/search.component.js
sorodrigo: This is an utility method, it doesn't depend on the instance, and its a pure function. I could move it outside of the class as a helper function, however I think that it makes more sense on leaving it inside because its totally dependent of the class implementation. There's no private methods in js, only static (class methods) and public.

frontend/scripts/react-components/shared/search.component.js
sorodrigo: :trollface: will give it a spin

frontend/scripts/react-components/shared/search.component.js
sorodrigo: Agree!

frontend/scripts/utils/getURLFromParams.js
sorodrigo: @nerik 👀

frontend/scripts/containers/shared/search-react.container.js
tiagojsag: I know. What I'm saying is that, instead of using a set string, we should compute it. Imagine that, in a few months, we get a "shipping company" column that has the same actors. We would need to create constants for every combination of it. bad for us

sorodrigo: true true, thanks I'll change it.

frontend/scripts/actions/tool.actions.js
tiagojsag: I see, and I agree we should change that behavior. However
- It seems to me that this is unrelated to this PR, so we should treat it in it's own PR
- We should review this more carefully because, as I pointed out above, this also affects other cases
- We should review these with Lena in our weekly meetings

frontend/scripts/react-components/shared/search.component.js
tiagojsag: :+1:

frontend/scripts/actions/tool.actions.js
sorodrigo: ok, can cherry pick this commits.

frontend/scripts/react-components/shared/search.component.js
sorodrigo: Now we rely on `lodash/deburr` so we support:
https://en.wikipedia.org/wiki/Latin-1_Supplement_(Unicode_block)#Character_table
https://en.wikipedia.org/wiki/Latin_Extended-A

Other alphabets not supported but don't think we need to.

sorodrigo:
@tiagojsag ready for re-reviewing!

frontend/scripts/actions/tool.actions.js
```diff
@@ -501,10 +501,35 @@ export function selectNodeFromGeoId(geoId) {
     const nodeId = getNodeIdFromGeoId(geoId, getState().tool.nodesDict, getState().tool.selectedColumnsIds[0]);
 
     // node not in visible Nodes ---> expand node (same behavior as search)
+    dispatch(selectExpandedNode(nodeId));
+  };
+}
+
+export function selectExpandedNode(nodeId) {
```
tiagojsag: I'm still not 100% happy with this. Let's discuss with JC

frontend/scripts/constants.js
```diff
@@ -1,3 +1,4 @@
+export const PROFILE_PAGES_WHITELIST = ['EXPORTER', 'IMPORTER', 'MUNICIPALITY'];
```
tiagojsag: I still believe we can get rid of this. Nodes have their own profile type, which is only set in nodes of these 3 types.

tiagojsag:
Left a few notes, and also there are conflicts

sorodrigo:
@tiagojsag I found an issue with the sankey behaviour change. Lets merge this branch as it is and address the behaviour change in another PR.

PR #87, opened 2017-11-22
Feat: dimensional selector

sorodrigo:
This PR includes:
- New country-commodity selector.
- New component `<DimensionalSelector />` that allows the user to select data in without an explicit order.
dimensional-selector


nerik:
:shipit:

nerik:
Few quick remarks before merging: 
Drop shadow is off and bottom text is too dark.

screen shot 2017-11-23 at 11 02 24

PR #110, opened 2017-11-23
add year dropdown to profile pages

nerik:
screen shot 2017-11-23 at 15 14 11


frontend/scripts/pages/profile-place.page.js
```diff
@@ -192,12 +188,11 @@ const _init = () => {
   const url = window.location.search;
   const urlParams = getURLParams(url);
   const nodeId = urlParams.nodeId;
-
-  const commodityDropdown = new Dropdown('commodity', _onSelect);
+  year = urlParams.year || 2015;
 
   commodityDropdown.setTitle(defaults.commodity);
```
tiagojsag: @nerik either this is not needed, or you deleted a few things that are

tiagojsag:
@nerik need your review on https://github.com/Vizzuality/trase/pull/110/commits/de63dd0ef403900a206dbca7bff93466a619c76a

tiagojsag:
A few things we need to look into, on actor profile:
- For some reason, the "top destination countries" charts are different for different years
- We might want to add the year to the "Comparing companies importing Soy from Brazil" chart title

Wasn't able to test the place chart

agnessa:
The way both the top destination countries and top sources charts work is that first we get the top nodes for the selected year, and then draw the charts across all years. This is why they're different, it could be different top nodes in different years. This makes a lot of sense to me, but because of the way the legend is rendered it is impossible to read those charts when the selected year != last available year. Should we fix it by calculating the "top nodes" for these 2 charts always for the last available year?

tiagojsag:
@juancarlosalonso can you do a quick review on that screenshot please? I think a hover effect wouldn't be bad, but you decide

PR #111, opened 2017-11-23
Mini sankey on place profile pages

nerik:
Replaces Top and Chord components with Mini Sankeys.

Currently uses dummy data, will need to be implemented on the backend, or data existing on the current endpoint must be converted on the frontend.

screen shot 2017-11-27 at 16 51 33

**TO DO**

- [ ] Load data from an endpoint (real or static json), instead of a hardcoded JS object
- [ ] Implement API to serve this data.
- [ ] Test that we show the same data that is currently on the chords (now removed)

### DIFF
--- a/frontend/html/profile-place.ejs
+++ b/frontend/html/profile-place.ejs
@@ -168,10 +168,10 @@
           </div>
         </section>
 
-        <section class="deforestation">
+        <section class="mini-sankey-container">
           <div class="row">
             <div class="small-12 columns">
-              <h3 class="title -small js-deforestation-title">-</h3>
+              <h3 class="title -small js-traders-title">-</h3>
               <div id="js-traders-sankey"></div>
             </div>
           </div>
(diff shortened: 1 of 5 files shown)


frontend/scripts/components/profiles/chord.component.js
```diff
@@ -1,111 +0,0 @@
-import { select as d3_select } from 'd3-selection';
```
tiagojsag: I'm wondering if it's not best to keep this around, properly documented, in case we need chords in the future. Experience has taught us that there's definitely some love out there for chord diagrams

sorodrigo: ☝️

nerik: I made a separated delete commit, that shouldn't be hard to revert.

nerik: As for documentation, it could be directly in the PR as suggested by @sorodrigo 
I'd ask @pjosh, as he's been the one writing that code (I don't think he's been deviating much from a standard D3 chord diagram though).

tiagojsag:
We can't really merge this until we have real data, because it removes real data in the process.

Also, @juancarlosalonso wouldn't it be a good idea to have a tooltip when hovering the flows, showing the actual amount of commodity traded?

juancarlosalonso:
yay to the tooltip
screen shot 2017-11-28 at 16 44 46

tiagojsag:
Added an optional URL param to show the new minisankeys intead of the old visualisations. This should allow us to merge this, and finish up the sankey API logic later.

PR #112, opened 2017-11-24
Feature/v3/columns

santostiago:
Added the endpoint for `api/v3/columns`

### DIFF
--- a/app/serializers/api/v3/columns/column_serializer.rb
+++ b/app/serializers/api/v3/columns/column_serializer.rb
@@ -1,0 +1,20 @@
+module Api
+  module V3
+    module Columns
+      class ColumnSerializer < ActiveModel::Serializer
+        attributes :id, :name, :position, :group, :is_default, :is_geo
+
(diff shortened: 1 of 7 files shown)


app/serializers/api/v3/columns/column_serializer.rb
```diff
@@ -0,0 +1,18 @@
+module Api
+  module V3
+    module Columns
+      class ColumnSerializer < ActiveModel::Serializer
+        attributes :id, :name, :position, :group, :is_default, :is_geo
+
+        attribute :profile_type do
```
tiagojsag: @santostiago you can join with the "profiles" table to load this from the DB, and not have it hardcoded

agnessa: That's true, I forgot about that - sorry @santostiago

santostiago:
Changed the route of columns from `/api/v3/columns?context_id=:context_id` to `/api/v3/contexts/:context_id/columns`

PR #134, opened 2017-11-30
Added the v3 endpoint for map groups

santostiago:
Created the v3 endpoint for map_groups.

There are two things missing (for which I'll need the help of @agnessa)
1 - The gold_master_urls file should have a way to add the `context_id` to the url that is not through params (I'll check where this is being used)
2 - I added the new fields of the `map_groups_mv` to the `schema_comments.yml` but those were not inserted in the `structure.sql`. I'll have to check where this is done.

Also, note that the test is not comparing the key `aggregateMethod` as this was always `nil` in v2.

### DIFF
--- a/app/controllers/api/v3/map_groups_controller.rb
+++ b/app/controllers/api/v3/map_groups_controller.rb
@@ -6,10 +6,17 @@ module Api
           where(context_id: @context.id).
           order('position ASC, id ASC')
 
-        dimensions = Api::V3::Readonly::MapAttribute.select('color_scale, id, bucket_3, bucket_5,
-map_attribute_group_id as group_id, is_default, years, name, attribute_type as type, unit, description,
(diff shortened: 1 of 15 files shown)


agnessa:
For the first issue, you can cheat like I did here: https://github.com/Vizzuality/trase/blob/255421ac3342918c425fa43146361799cada23e8/spec/support/gold_master_urls.yml

(splitting into 2 urls by context)

The comments are inserted when you run `bundle exec rake db:revamp:doc:sql`

agnessa:
Could you also update the swagger spec?

santostiago:
@agnessa done.

santostiago:
@agnessa rebased.

agnessa:
I think this produces duplicates? For example for `/api/v3/contexts/1/map_groups?context_id=1` I can see "Production of soy in Tn" twice. This gets reported as a difference when you run the gold master tests as well.

PR #135, opened 2017-11-30
Added the v3 endpoint for newsleters_subscriptions

santostiago:
This is a very simple feature, as it's just copying it from V2.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -15,6 +15,7 @@ Rails.application.routes.draw do
         resources :columns, only: [:index]
       end
       get '/get_all_nodes', to: 'nodes#get_all_nodes'
+      resources :newsletter_subscriptions, only: [:create]
     end
     namespace :v2 do
       resources :geo_id, only: :index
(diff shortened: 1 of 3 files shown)


config/routes.rb
```diff
@@ -15,6 +15,7 @@
         resources :columns, only: [:index]
       end
       get '/get_all_nodes', to: 'nodes#get_all_nodes'
+      resources :newsletter_subscriptions, only: [:create]
```
agnessa: maybe don't remove it from v2 just yet, so that we don't break anything in the frontend?

santostiago: Yes, I know. I didn't remove it. The routes were duplicated, so I just removed one of them.
It still exists in line 24.

tiagojsag:
no tests??????

santostiago:
@tiagojsag I've done the tests, but I removed them as I found them pointless.
For the test to pass, there should be a valid mail chimp API key. And it would actually add the email provided to the mailing list (as this is the only thing this endpoint does). So I thought it would be better to remove the test.
@agnessa I've updated the swagger file.

PR #136, opened 2017-12-01
Add v3 api enpoint logic to frontend

tiagojsag:


frontend/.eslintrc.js
```diff
@@ -41,6 +41,7 @@ module.exports = {
       NODE_ENV_DEV: true,
       API_V1_URL: true,
       API_V2_URL: true,
+      API_V3_URL: true,
```
agnessa: please add to `.env.sample`

tiagojsag: done

PR #137, opened 2017-12-01
First version of the V3 endpoints for Nodes

santostiago:
This feature is not complete as I wasn't able to test it.
When including `brazil contexts` I get an error creating the schemaRevamp as follows:
```
NoMethodError: undefined method `[]' for nil:NilClass
./lib/schema_revamp.rb:633:in `block in insert_chart'
```

@agnessa would you mind taking a lot at it?

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -9,6 +9,7 @@ AllCops:
     - 'config/initializers/*'
     - 'config/*'
     - 'frontend/node_modules/**/*'
+    - lib/schema_revamp.rb
 
 Bundler/OrderedGems:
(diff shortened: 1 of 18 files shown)


agnessa:
Fixed the issue in SchemaRevamp. Please ensure you also run the gold master tests, they sometimes are more helpful in understanding the source of differences. And make sure you have the commit hooks enabled for code style checks.

agnessa:
I'm still working on this one, have one problem remaining to fix.

agnessa:
I think this is now good to merge.

PR #142, opened 2017-12-07
Styling details and error handling on profile pages

### LINKED ISSUE #117: Handle 5xx errors on profile pages
If something goes wrong with the API call, the profile page does not inform the user

tiagojsag:
Fixes #117 

- When loading data on profile search page, it now shows a message on failure
- On the same page, the CSS is no longer broken while data is loading
- On profile pages (actor and place) an error message is displayed in case data fails to load

### DIFF
--- a/frontend/scripts/pages/profile-actor.page.js
+++ b/frontend/scripts/pages/profile-actor.page.js
@@ -331,7 +331,7 @@ const _init = ()  => {
       if (response.ok) {
         return response.json();
       }
-      _showErrorMessage(response.statusText);
+      throw new Error(response.statusText);
     })
     .then((result) => {
       if (!result) return;
@@ -360,9 +360,7 @@ const _init = ()  => {
 
       _build(data, nodeId);
     })
-    .catch(
-      (reason) => _showErrorMessage(reason.message)
-    );
+    .catch((reason) => _showErrorMessage(reason.message));
 
   const nav = new Nav();
   print = nav.print;
--- a/frontend/scripts/pages/profile-actor.page.js
+++ b/frontend/scripts/pages/profile-actor.page.js
@@ -276,12 +276,15 @@ const _setEventListeners = () => {
   smoothScroll(document.querySelectorAll('.js-link-profile'));
 };
 
-const _showErrorMessage = () => {
+const _showErrorMessage = (message = null) => {
   const el = document.querySelector('.l-profile-actor');
   el.classList.add('-error');
   document.querySelector('.js-loading').classList.add('is-hidden');
   el.querySelector('.js-wrap').classList.add('is-hidden');
   el.querySelector('.js-error-message').classList.remove('is-hidden');
+  if (message !== null) {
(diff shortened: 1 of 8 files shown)


frontend/scripts/pages/profile-actor.page.js
```diff
@@ -276,12 +276,15 @@ const _setEventListeners = () => {
...
 
-const _showErrorMessage = () => {
+const _showErrorMessage = (message = null) => {
   const el = document.querySelector('.l-profile-actor');
   el.classList.add('-error');
   document.querySelector('.js-loading').classList.add('is-hidden');
   el.querySelector('.js-wrap').classList.add('is-hidden');
   el.querySelector('.js-error-message').classList.remove('is-hidden');
+  if (message !== null) {
+    el.querySelector('.message').innerHTML = message;
```
sorodrigo: Its a good practice to add js specific classes for DOM operations i.e. `.js-message`.

frontend/scripts/pages/profile-place.page.js
```diff
@@ -178,12 +178,16 @@ const _setEventListeners = () => {
...
 
-const _showErrorMessage = () => {
+const _showErrorMessage = (message = null) => {
   const el = document.querySelector('.l-profile-place');
   document.querySelector('.js-loading').classList.add('is-hidden');
   el.classList.add('-error');
   el.querySelector('.js-wrap').classList.add('is-hidden');
   el.querySelector('.js-error-message').classList.remove('is-hidden');
+  if (message !== null) {
+    el.querySelector('.message').innerHTML = message;
```
sorodrigo: same

frontend/styles/layouts/l-profiles.scss
```diff
@@ -41,6 +53,20 @@
         font-size: $font-size-medium;
       }
 
+      > .c-search > .js-search-input {
```
sorodrigo: Avoid using `js` prefixed classes for styling.

frontend/scripts/pages/profile-actor.page.js
```diff
@@ -325,15 +328,12 @@ const _init = ()  => {
...
-        return null;
-      }
-
-      if (response.status === 200) {
+      if (response.ok) {
         return response.json();
       }
-    })
+      _showErrorMessage(response.statusText);
+    }, (reason) => _showErrorMessage(reason.message)
```
sorodrigo: Nifty af! I didn't know of the existence of Promise.then second argument. 🙌 
As a side note, I prefer using `.catch` for handling errors but this is perfectly fine.

frontend/scripts/pages/profile-actor.page.js
```diff
@@ -325,15 +328,12 @@ const _init = ()  => {
...
+    )
     .then((result) => {
       if (!result) return;
```
sorodrigo: On a second thought, if we use catch, we don't have to early return in the next then.

tiagojsag:
I think i got them all. let me know @sorodrigo

Issue #102, opened 2017-11-23
Site feedback design for userreport.com widget


tiagojsag:
You can see it live on staging. We are very limited in what we can customize - color, left/right side, and I don't know if much more. Still, let us know if you'd like to make any change to it. I personally think it looks pretty bad on the map page, but that's just me.

juancarlosalonso:
I think we should move it to the bottom right corner of the page. 
`bottom:12px;
right:0px; 
`
<img width="1280" alt="screen shot 2017-11-27 at 10 26 39" src="https://user-images.githubusercontent.com/20836216/33259555-8624e776-d35d-11e7-8d51-ec2ab770468b.png">

tiagojsag:
https://userreport.com/test/fb-examples/#2

https://helpdesk.userreport.com/hc/en-us/articles/209410026-Disabling-the-feedback-tab

Seems like it's possible to implement custom buttons. However, we want to spend very little time on this, so if you want to do a custom design, please check with @sorodrigo to make sure we have something simple and easy to implement.

juancarlosalonso:
I'm happy with the way it works now. Would be great to change the icon, though.
<img width="363" alt="screen shot 2017-12-01 at 10 52 04" src="https://user-images.githubusercontent.com/20836216/33477365-b6549330-d685-11e7-83c3-ecc805dc63ff.png">

tiagojsag:
yeah, the current icon is rather creepy :P

juancarlosalonso:
the new icon is here: https://www.dropbox.com/s/rpfpko9nxjgtllr/feedback.svg?dl=0

PR #143, opened 2017-12-07
Feat: feedback button

sorodrigo:
This PR includes:
- New feedback button. Its easier to replace the current button with a custom one than to try to hack the styles of the existing.
- Creates new static include file `feedback.ejs` this will be included in all pages.

_Note: Given that icons are not included in all pages (i.e. FAQs) the svg used in the feedback button is placed inline. Same with the button styles (included in the base styles)._

Fixes: #102


sorodrigo:
Here's the instructions for disabling the default button: https://helpdesk.userreport.com/hc/en-us/articles/209410026-Disabling-the-feedback-tab

tiagojsag:
Let's review the design with @juancarlosalonso

PR #146, opened 2017-12-07
Feature/get place attributes v3

agnessa:
`/contexts/:context_id/nodes/:node_id/place`

To make the old and new response comparable I had to cheat. Since we wanted to remove the "get value from closest year when not available for current year" feature, I removed it from the original response.

It seems this is two times faster than the original endpoint.


app/models/api/v3/flows_result.rb
```diff
@@ -0,0 +1,83 @@
+module Api
```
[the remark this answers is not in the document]
agnessa: yes, and it had been moved out of there, not sure exactly how it made it's way back <facepalm>

app/models/api/v3/node_type.rb
```diff
@@ -3,6 +3,30 @@ module V3
...
+        NodeTypeName::STATE
+      ].freeze
+
+      def self.node_index_for_name(context, node_type_name)
+        zero_based_idx = ContextNode.
+          joins(:node_type).
+          where(context_id: context.id).
+          where('node_types.node_type' => node_type_name).
+          pluck(:column_position).first
+        zero_based_idx && zero_based_idx + 1 || 0
```
santostiago: Isn't this the same as `zero_based_idx + 1 || 0`?

agnessa: I think `zero_based_idx` can be nil, in which case `+` would result in "undefined method + for nil"

PR #156, opened 2017-12-12
Feature/node attributes v3

agnessa:
/contexts/:context_id/nodes/attributes


tiagojsag:
could you update the frontend config too, please?

agnessa:
I think the frontend was already updated, as in the url to use - or you mean sth different? I just rebased.

tiagojsag:
oh, no, must have missed that, sorry

tiagojsag:
should we add response structure tests as part of this PR?

agnessa:
would be good, but I won't have time to do it before Christmas; there are more endpoints that are missing these tests, so an alternative would be to add all the missing ones in a separate PR.

Issue #85, opened 2017-11-22
Add info on national/subnational data availabilty


sorodrigo:
Currently neither `context_node_types` nor `node_types` have info about what kind of data is available. A `hasSubnationalLevelData` attribute could be added to either of them.

![screen shot 2017-11-22 at 12 38 35](https://user-images.githubusercontent.com/12124839/33126860-6327cc64-cf86-11e7-86a6-9da9c83ac921.png)

PR #159, opened 2017-12-14
Add is_subnational flag to api

tiagojsag:
fixes https://github.com/Vizzuality/trase/issues/85


tiagojsag:
@sorodrigo I did a significant refactor of your country-commodity selector. I sacrificed a bit of reusability in favor of (what I think is) simpler code, that is easier to understand. Not that I think your initial approach was bad, just that I don't think we'll end up reusing those components that much

let me know

sorodrigo:
I fixed some warnings. Looks good!

PR #162, opened 2017-12-18
Feature/get actor attributes v3

agnessa:
/contexts/:context_id/nodes/:node_id/actor?year=:year


app/services/api/v3/actor_node/basic_attributes.rb
```diff
@@ -0,0 +1,214 @@
...
+          text = summary_of_total_trade_volume('importer')
+          text += summary_of_sources('importer')
+          text += summary_of_destinations('importer')
+          text
+        end
+
+        def summary_of_total_trade_volume(profile_type)
+          initialize_trade_volume_for_summary
+          text = "#{@node.name.humanize} was the \
+#{@trade_total_rank_in_country_formatted}largest #{profile_type} of soy \
```
tiagojsag: there's a space missing, and there's a bug that's causing `@trade_total_rank_in_country_formatted` to not render anything

app/services/api/v3/actor_node/exporting_companies_plot.rb
```diff
@@ -0,0 +1,117 @@
...
+                attribute_hash.slice(:name, :unit)
+              end,
+              companies: exports
+            }
+          }
+        end
+
+        private
+
+        def attributes_list
```
tiagojsag: We should pass this info on to Clement, so he knows these attributes are required.
Also, can't we load names and units from the DB? this info should be there already.

app/services/api/v3/actor_node/response_builder.rb
```diff
@@ -0,0 +1,49 @@
...
+          )
+          sustainability = Api::V3::ActorNode::SustainabilityTable.new(
+            @context, @year, @node
+          ).call
+
+          exporting_companies = Api::V3::ActorNode::ExportingCompaniesPlot.new(
+            @context, @year, @node
+          ).call
+
+          @basic_attributes.attributes.
```
tiagojsag: <3

app/services/api/v3/actor_node/sustainability_table.rb
```diff
@@ -0,0 +1,126 @@
...
+            is_total: true,
+            values: @attributes.map do |attribute_hash|
+              if group_totals_hash[attribute_hash[:attribute_name]]
+                {value: group_totals_hash[attribute_hash[:attribute_name]]}
+              end
+            end
+          }
+        end
+
+        def attributes_list
```
tiagojsag: same as before, let's document that these are required

frontend/scripts/utils/getURLFromParams.js
```diff
@@ -24,19 +24,15 @@ const API_ENDPOINTS = {
   [GET_ALL_NODES]: { api: 3, endpoint: '/contexts/$context_id$/nodes' },
   [GET_NODE_ATTRIBUTES]: { api: 3, endpoint: '/contexts/$context_id$/nodes/attributes' },
   [GET_MAP_BASE_DATA]: { api: 3, endpoint: '/contexts/$context_id$/map_groups' },
-  [GET_PLACE_FACTSHEET]: { api: 3, endpoint: '/contexts/$context_id$/nodes/$node_id$/place' },
   [GET_INDICATORS]: { api: 3, endpoint: '/contexts/$context_id$/download_attributes' },
+  [GET_PLACE_FACTSHEET]: { api: 3, endpoint: '/contexts/$context_id$/nodes/$node_id$/place', mock: 'mocks/v1_get_place_node_attributes.json' },
```
tiagojsag: these 2 lines are too long

app/services/api/v3/actor_node/basic_attributes.rb
agnessa: I think the space is not missing here - it's in line 83. It renders nothing when the rank is 1 - that's a feature.

app/services/api/v3/actor_node/exporting_companies_plot.rb
agnessa: This info is not in the db. We have designed tables where these lists of attributes could be kept, and I guess it would be possible to store them there, but without a tool to update them this has little value. We've discussed not implementing the fully dynamic profile pages with configurable attributes in this phase, so I kept it as is. Good point about providing Clement with a full list of currently used attributes.

app/services/api/v3/actor_node/response_builder.rb
agnessa: naming things is a hard job

app/services/api/v3/actor_node/sustainability_table.rb
agnessa: documented in the "Profile definitions" spreadsheet, Clement informed about update to that document.

app/services/api/v3/actor_node/basic_attributes.rb
tiagojsag: I know, but that "feature" is present on all actors :P Just double checked with Amaggi 2015. So it seems my comment should have been on a different line, but the bug is there.

agnessa: yes you are right, shame on me! 🔔 🔔 🔔

PR #167, opened 2017-12-19
Feature/download attributes v3

agnessa:
/contexts/:context_id/download_attributes


spec/controllers/api/v2/indicators_controller_spec.rb
```diff
@@ -7,7 +7,7 @@
     it 'assigns indicators relevant to the given context' do
       get :index, params: {context_id: context.id}
       expect(assigns(:context)).to eq(context)
-      expect(assigns(:indicators)).to match_array([forest_500, deforestation_v2, zero_deforestation])
+      expect(assigns(:indicators)).to match_array([deforestation_v2, zero_deforestation])
```
tiagojsag: this caught my eye. any reason why we don't like forest 500 anymore?

agnessa: That's because currently it's not on the list of attributes available for download.

PR #170, opened 2017-12-19
Reactify Top component

tiagojsag:


tiagojsag:
moved into https://github.com/Vizzuality/trase/pull/174

PR #172, opened 2017-12-19
Feat: add correct page breaks

sorodrigo:
This PR includes:
- The print styles were missing some proper _page-breaks_, this PR adds them so that everything fits nicely.

[ ] **TODO:**`The chord diagrams won't fit in an A4 page with their current dimension, we need to shrink them when print query-param is true` .
[ ] **TODO:**`The choropleth legends colors arent showing up (just like background isn't either)`.
[ ] **TODO:**`The map outline should be white when print query-param is true`.


tiagojsag:
please fix conflicts, otherwise LGTM

sorodrigo:
should we add the TODOs to the backlog?

tiagojsag:
yes

PR #188, opened 2018-01-08
Feature/download v3

agnessa:
/api/v3/contexts/:context_id/download.[csv|json]

### DIFF
--- a/app/models/api/v3/download_version.rb
+++ b/app/models/api/v3/download_version.rb
@@ -1,6 +1,12 @@
 module Api
   module V3
     class DownloadVersion < BaseModel
+      def self.current_version_symbol(context)
+        current_version = where(is_current: true, context_id: context.id).
+          order('created_at DESC').
+          first
+        current_version&.symbol
+      end
     end
   end
 end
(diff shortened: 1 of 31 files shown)


app/models/api/v3/download_version.rb
```diff
@@ -1,6 +1,12 @@
 module Api
   module V3
     class DownloadVersion < BaseModel
+      def self.current_version_symbol(context)
+        current_version = where(is_current: true, context_id: context.id).
+          order('created_at DESC').
```
tiagojsag: image

I'm getting v1.0 for Brazil soy. We should probably add a 2nd "order by" clause, per symbol maybe.

agnessa: I think that's because there was a bug in the SchemaRevamp.copy logic, where `created_at` was set to `NOW()` at the time of copying rather than copying the existing `created_at` value. That's also fixed in this PR, but requires running the copy task again.

agnessa: If we think `created_at` is not reliable, what would be good I guess is a constraint to allow only one row to have `is_current` set.

tiagojsag: that would make sense, yes.

agnessa: I added the constraint to our database, I'll also recommend adding it to main db.

Issue #128, opened 2017-11-27
Add tooltips to contextual layers


tiagojsag:
Need to:

- [x] Get the actual tooltips for each layer from SEI
- [x] Implement the data columns + update API to serve this data
- [x] Show it on the FE

PR #194, opened 2018-01-11
Load contextual layers info from API; rename map_groups to map_layers

tiagojsag:
Fixes https://github.com/Vizzuality/trase/issues/128


db/structure.sql
```diff
@@ -128,7 +128,7 @@ CREATE FUNCTION get_trader_sum(trader_id integer, year_ integer) RETURNS double
           FROM flows
           WHERE trader_id = ANY(path)
             AND year = year_
-            AND context_id = 1);
```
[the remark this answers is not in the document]
tiagojsag: no, probably some leftover from db:migrate changes to structure.sql. will revert. I'll leave the cleaning up for later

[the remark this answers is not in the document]
tiagojsag: oh, just noticed that this is just my IDE removing trailling spaces

tiagojsag:
@agnessa @sorodrigo could you review this, please? thanks

app/models/api/v3/contextual_layer.rb
```diff
@@ -3,6 +3,8 @@ module V3
     class ContextualLayer < BaseModel
       belongs_to :context
 
+      has_one :carto_layer
```
agnessa: this should be a `has_many`, allowing to have different carto layers for different years

tiagojsag:
@agnessa should be ready for final review

app/models/api/v3/contextual_layer.rb
```diff
@@ -3,6 +3,8 @@ module V3
     class ContextualLayer < BaseModel
       belongs_to :context
 
+      has_many :carto_layer
```
agnessa: Let's make the relation name plural please, this is strange :) We also need to consider how the front-end will handle the fact that different years could have different layers. I guess we could either parametrise the call to `map_layers` with year, and then we could filter out only relevant layers, or if we're serving all layers we need to let front-end know which year range each is for - I believe the `years` column is currently not serialised.

tiagojsag: relationship renamed.

As handling different layers for different years: that will require a bit more thought, but maybe that can be done on a separate PR

Issue #178, opened 2018-01-04
Landing page: implement backend support for new content structure


tiagojsag:
- [x] "Category" field (insights, news, etc) for the already existing stories
- [x] New "Testimonials" endpoint: with "quote", "author name", "author title" and "author pic"
- [x] On tweets, return avatar url/img
- [x] "Starred" flag on stories: limit to 1 (reuse highlighted). When a admin highlights a post, de-highlight any previously highlighted posts
- [x] potentially new types of content for "insights" (https://basecamp.com/1756858/projects/12498794/messages/73758553)
- [x] Remove "content" field from post content type
- [x] Remove logic that loads different images based on highlighted flag
- [x] Add a "type of link" select, with "file" or "webpage"

tiagojsag:
- "Category" field (insights, news, etc) for the already existing stories
- New "Testimonials" endpoint: with "quote", "author name", "author title" and "author pic"
- On tweets, return avatar url/img
- "Starred" flag on stories: limit to 1

tiagojsag:
"Testimonials" done on https://github.com/Vizzuality/trase/pull/210

agnessa:
frontend https://github.com/Vizzuality/trase/issues/212

PR #210, opened 2018-01-17
Add testimonials model to content api

tiagojsag:
Partially addresses https://github.com/Vizzuality/trase/issues/178


tiagojsag:
Added model validators.

`ApplicationController.helpers.asset_url` automatically adds `action_controller.asset_host` config param when it's present, which it is for staging and prod, just not for dev. So we will get the full path on those envs

tiagojsag:
@agnessa ready for re-review

PR #214, opened 2018-01-17
Fixed leftover part of query building using old style identifier names

agnessa:


tiagojsag:
LGTM once travis is green

PR #220, opened 2018-01-18
Minor refactor on slider component

tiagojsag:


sorodrigo:
@tiagojsag I removed `<React.Fragment></React.Fragment>` this is only necessary if you want to render two or more elements at root level

Issue #103, opened 2017-11-23
New chord/mini-sankey on the profile pages


tiagojsag:
Since we want to split up profile page responses, let's use a mock, standalone json for this data, if using the existing response is not possible. Might be worth chatting with Agnieszka about the json structure

nerik:
Data structure currently used on the FE


```
const tradersSankeyData = {
    name: 'QUERÊNCIA',
    targetNodes: [{
      id: 2,
      name: 'Other',
      isAggregated: true,
      height: 0.2,
      // isDomesticConsumption: null,
    },{
      id: 588,
      name: 'Cargill',
      isUnknown: null,
      height: 0.4,
      profileType: 'actor'
    },{
      id: 588,
      name: 'Cargill',
      isUnknown: null,
      height: 0.2,
      profileType: 'actor'
    },{
      id: 588,
      name: 'Cargill',
      isUnknown: null,
      height: 0.1,
      profileType: 'actor'
    },{
      id: 588,
      name: 'Cargill',
      isUnknown: null,
      height: 0.03,
      profileType: 'actor'
    },{
      id: 588,
      name: 'Cargill',
      isUnknown: null,
      height: 0.02,
      profileType: 'actor'
    },{
      id: 588,
      name: 'Cargill',
      isUnknown: null,
      height: 0.05,
      profileType: 'actor'
    }]
  };
```

tiagojsag:
Currently WIP, missing hover effect

PR #228, opened 2018-01-18
Feature/complete minisankey work

tiagojsag:
https://github.com/Vizzuality/trase/issues/103


tiagojsag:
Good to review, will fix tests tomorrow

tiagojsag:
Seems like we have an issue on some cases, where the paths are not rendered correctly

http://localhost:8081/profile-place?nodeId=653&year=2015

image

frontend/scripts/pages/profile-place.page.jsx
```diff
@@ -160,156 +160,47 @@ const _build = (data, { year, showMiniSankey }, store) => {
...
-    };
-
-    render(
-      <MiniSankey
-        data={tradersSankeyData}
-        targetLink="actor"
-      />,
-      document.getElementById('js-traders-sankey')
-    );
+  document.querySelectorAll('.mini-sankey-container')
```
sorodrigo: does this work? 👀

tiagojsag: seems to, but it's actually not needed, so I'll remove it

sorodrigo: ok I'm impressed I didn't know this. However, for the future => https://developer.mozilla.org/en-US/docs/Web/API/NodeList/forEach

It's not supported on IE11 :(

tiagojsag:
ready for re-review team

PR #229, opened 2018-01-19
Feat: new about

sorodrigo:
This PR includes:
- About page MVP. There's still plenty to do like:
    - About sections (Team, Partners, Funders).
    - Rest of static pages.
    - Nav sidebar subsections (with separator).

But I figured better to approach them in other PRs and not bloat this one (already 29 files long).


frontend/html/about-old.ejs
```diff
@@ -0,0 +1,237 @@
+<div class="loading-wrapper">
```
tiagojsag: is this file needed?

frontend/scripts/pages/about-old.page.jsx
```diff
@@ -0,0 +1,93 @@
+/* eslint-disable no-new */
+import AboutMarkup from 'html/about-old.ejs';
```
tiagojsag: and this?

frontend/scripts/index.ejs
```diff
@@ -6,6 +6,7 @@
 <%= htmlWebpackPlugin.options.icons %>
 
 <div id="app-root-container"></div>
+<div id="app-portal-container"></div>
```
tiagojsag: Don't really see the point of using a portal here. Are you sure this is the best option?

frontend/scripts/react-components/shared/portal.component.jsx
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
```
tiagojsag: is it me, or this is not actually used?

frontend/scripts/router.js
```diff
@@ -65,35 +70,45 @@ const routes = {
...
     path: '/FAQ',
-    page: 'FAQ',
-    extension: 'jsx'
+    page: 'static-content',
+    extension: 'jsx',
+    thunk: dispatchThunks(getPageStaticContent)
   },
   [NOT_FOUND]: {
-    path: '/404'
+    path: '/404',
```
tiagojsag: Please create a separate issue for us to test if this is actually returning an 404 http code

frontend/scripts/react-components/static-content/static-content.component.jsx
```diff
@@ -0,0 +1,54 @@
...
+import NavSidebar from './nav-sidebar.component';
+
+class StaticContent extends React.PureComponent {
+  render() {
+    const { links = [], content = '', children, notFound } = this.props;
+    const MarkdownContainer = p => (<div className="markdown-content">{p.children}</div>);
+    return (
+      <div className="c-static-content">
+        <Hero className="-read-only" />
+        {!notFound &&
```
tiagojsag: Kudos for not using trenary operator here!!!!

tiagojsag: This should probably be two separate components.

frontend/scripts/react-components/static-content/static-content.actions.js
```diff
@@ -0,0 +1,32 @@
...
+export const getStaticContentFilename = location => camelCase(
+  `${location.type} ${(location.payload.section || '')}`
+);
+
+export const getStaticContent = () => (dispatch, getState) => {
+  const { location, staticContent } = getState();
+  const filename = getStaticContentFilename(location);
+  const notFound = location.type === NOT_FOUND;
+  if (
+    typeof staticContent.markdown[filename] === 'undefined' &&
```
tiagojsag: wouldn't `staticContent.markdown[filename] === undefined` be equivalent? just asking, no need to change if they're the same

frontend/scripts/react-components/static-content/static-content.actions.js
```diff
@@ -0,0 +1,32 @@
...
+    !notFound
+  ) {
+    const url = `/static-content/${filename}.md`;
```
tiagojsag: why not use my awesome beautiful getUrl class? https://github.com/Vizzuality/trase/blob/94f3270dfc54f9b79cc9fbfe7c1cc802aac5e93f/frontend/scripts/utils/getURLFromParams.js

frontend/scripts/react-components/static-content/static-content-links.js
```diff
@@ -0,0 +1,42 @@
+
+export default {
```
tiagojsag: I'm wondering if moving this into router.js makes sense. Seems like the first place us-in-6-months will look for when we need to add more pages. What do you think?

sorodrigo: I actually struggled deciding where to put this. If we put it in the router, we should split that file.

sorodrigo: Actually I decided against this, the links have more sense in the `static-content` context. If you take them up to the router it will be confusing. Besides, it's currently a relative import so following the breadcrumbs won't be hard. 🍪

frontend/scripts/react-components/static-content/static-content.component.jsx
sorodrigo: I knew you'd like it. Food for thought. https://medium.com/javascript-scene/nested-ternaries-are-great-361bddd0f340

frontend/scripts/react-components/static-content/static-content-links.js
sorodrigo: In the end when the new nav, refactoring this made more sense. I made this change on the new nav PR

frontend/scripts/react-components/static-content/static-content.component.jsx
tiagojsag: Hum I don't really agree with that article. The author basically advocates that trenary operators are great because they are less verbose and clutered than if/else/return statements, which they are, but that does not make it easier for humans to understand and debug. Humans many times need help from those "noisy" operators, otherwise we'd all be programming in https://en.wikipedia.org/wiki/Whitespace_(programming_language)

tiagojsag:
Seems we have some homepage css regressions introduced in this PR :/ otherwise LGTM

frontend/scripts/react-components/static-content/static-content.component.jsx
sorodrigo: hahaha 👍

sorodrigo:
ready

PR #232, opened 2018-01-22
Reactify profile search page

tiagojsag:


tiagojsag:
@sorodrigo ready to review

frontend/scripts/react-components/profile-search/profile-search.reducer.js
```diff
@@ -0,0 +1,24 @@
...
+  SET_PROFILE_SEARCH_ERROR_MESSAGE,
+  SET_PROFILE_SEARCH_NODES
+} from 'react-components/profile-search/profile-search.actions';
+
+const initialState = {
+  nodes: [],
+  errorMessage: null
+};
+
+const homeReducer = {
```
sorodrigo: dead kitty. :(

frontend/scripts/react-components/profile-search/profile-search.reducer.js
```diff
@@ -0,0 +1,24 @@
...
+  [SET_PROFILE_SEARCH_NODES](state, action) {
+    const { nodes } = action.payload;
+    return { ...state, nodes: clone(nodes) };
```
sorodrigo: In the `homeReducer` I use clone because the same reducer saves arrays and objects, and this way I don't have to worry about what I'm receiving. In this case, it doesn't really matter, but we could trim a (perhaps negligible) dependency using `[...nodes]` syntax.

frontend/scripts/react-components/profile-search/profile-search-box.container.js
```diff
@@ -0,0 +1,16 @@
+import { connect } from 'react-redux';
+import ProfileSearchBox from 'react-components/profile-search/profile-search-box.component';
+import { bindActionCreators } from 'redux';
+import { goToNodeProfilePage } from 'react-components/profile-search/profile-search.actions';
+
+function mapStateToProps(state) {
+  return {
+    nodes: state.profileSearch.nodes
```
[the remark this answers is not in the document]
tiagojsag: we need this, at least for now, as the tool's node list depends on the context, but not this one.

frontend/scripts/react-components/profile-search/profile-search.reducer.js
tiagojsag: kitty revived

frontend/scripts/react-components/profile-search/profile-search.reducer.js
tiagojsag: fixed, thanks for the tip :+1:

sorodrigo:
This is ready!

sorodrigo:
Made a change to your change, the background should be white always and the color of the placeholder should change. Added some 🎩  ✨  🐰  also. Tested on firefox quantum and replicted the previous issue, safe to say it's fixed. Merging.

PR #233, opened 2018-01-22
Feature/fdw connection to trase main

agnessa:
This adds a new importer to fetch data directly from main database and update in Trase. The importer is still being tested against the live database, so in that way it's still work in progress, but since this is quite a big chunk of work I thought it makes sense to bring it to your attention already and get feedback.

The logic of the import script is described in the README.


agnessa:
I think I'm done with this!

agnessa:
I'll check again using the new db connection.

tiagojsag:
Merging just because I don't want to review aaaaallllllll of this again. I'm sure there are improvements we can make, but let's make them in a separate, smaller PR.

PR #243, opened 2018-01-29
Feat: home flows animation

sorodrigo:
This PR includes:
- React component `<AnimatedFlows />` that wraps the flows animation d3 component.

Kudos to @juancarlosalonso the final result is amazing!

Please merge after #242


frontend/package.json
```diff
@@ -82,6 +83,7 @@
     "query-string": "^5.0.1",
     "react": "^16.2.0",
     "react-dom": "^16.2.0",
+    "react-faux-dom": "^4.1.0",
```
tiagojsag: I think this is not used, and can be removed, right?

sorodrigo: si thly

PR #246, opened 2018-01-31
Feat: perf test

sorodrigo:
This PR includes new performance testing tools:
- Why Did You Update (piggybacks on React to console log when a component is rendering without need).

To enable the perf test a new env variable should be added: 
```sh
PERF_TEST=true
```


tiagojsag:
@sorodrigo  please add the var to .env.sample and to the README

PR #254, opened 2018-02-01
Sanitize tool reducer

tiagojsag:
- Added complete default state
- Removed 1 state value and 1 aux function that were not needed
- Fixed some issues resulting from new default state (validations relying on undefined values)
- Optimized lodash usage in some places
- Fixed issue where recolor by, resize by and biome filter were not saved on the URL

Will add proper documentation to default state at a later stage


sorodrigo:
I see why would you use `isEmpty`, but can't we just set the initialState to `null`? I think it's a more intuitive approach. Perhaps I'm missing something, open for discussion.

Also, let's merge after #232 to take advantage of the new eslint config.

tiagojsag:
The thing is that, before, we had a lot of undefined state by default. Now we use nulls, but because state was being compared with === undefined, things broke. So, to avoid that, isEmpty is our friend <3

sorodrigo:
But `isEmpty` is used to check if `Object.keys({}).length === 0`. If the state is null or undefined a simple `!!variable` will do.

tiagojsag:
Now I'm lost, sorry. Could you specify to which part of the code you are referring to?

Issue #216, opened 2018-01-17
Not possible to change basemap


tiagojsag:
Seems that the basemap radio buttons on the map page are always greyed out, not allowing the user to switch basemaps. The logic should be that the basemap should be set to "default" if there are chorpleth layers selected, and the user should not be able to change basemap while there are choropleth layers enabled

Test:
- The user should be able to change between different basemaps
- Test reloading page with and without chorpleth layers selected, and see if logic above is respected
- Test on contexts with and without choropleth layers

PR #257, opened 2018-02-01
Fix issue preventing selection of basemap

tiagojsag:
Fixes https://github.com/Vizzuality/trase/issues/216


sorodrigo:
Let's merge this after #232 to take advantage of the new eslint config

PR #262, opened 2018-02-02
Feature/yellow tables admin interface

agnessa:
Adds admin interface for all "yellow tables" except for the ones related to charts. We're not using those tables at all at the moment and with potential changes in the next phase I thought it best not avoid possibly unnecessary functionality.

A few of the AR models share common functionality related to validating and managing relationships with  respective _inds / _quals / _quants tables; this applies to DownloadAttribute, MapAttribute, ResizeByAttribute and RecolorByAttribute.

The admin interface takes advantage of column comments in db/schema_comments.yml to display hints.

Where changes to the yellow tables require recalculation of materialized views there's an after_commit hook. One situation which is not handled is refreshing the materialized view which serves downloads; this needs to happen in a background job, and while that logic is ready it is commented out, because it still needs changes in respect to managing sidekiq queues - currently it would have been possible to start multiple refreshes at the same time. Ideally only one should be processed at a time and this capability will be added in a separate PR.

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -7,6 +7,8 @@ engines:
     enabled: true
   duplication:
     enabled: true
+    exclude_paths:
+      - app/admin/api/v3/
     config:
       languages:
       - ruby:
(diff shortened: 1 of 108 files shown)


agnessa:
The main nav menus wrapped on my screen, so I reduced their number splitting settings into General, Sankey, Map and Downloads related.

PR #263, opened 2018-02-05
feat: includes prop-types in redux reducers 🎤

tiagojsag:
- Restructured the reducers to match the new structure
- removed that awful actions/index.js file, and spread its content across the different files
- added validation to state
- fixed whatever the state validation turned up


frontend/config/webpack.config.dev.js
```diff
@@ -16,10 +16,8 @@ module.exports = merge(webpackBaseConfig, {
     host: '0.0.0.0',
     historyApiFallback: true
   },
-  plugins: [
-    new webpack.NamedModulesPlugin()
-  ],
+  plugins: [new webpack.NamedModulesPlugin()],
   performance: {
-    hints: 'warning'
+    hints: false
```
tiagojsag: not related, but this is annoying as hell, so I disabled it

sorodrigo: fine

PR #264, opened 2018-02-05
Prettier fy code

tiagojsag:


frontend/.prettierignore
```diff
@@ -1,3 +1,4 @@
-node_modules
```
tiagojsag: https://prettier.io/docs/en/cli.html#with-node-modules

frontend/scripts/actions/app.actions.js
```diff
@@ -8,7 +8,7 @@ export function resize() {
 }
 
 export function toggleMap() {
-  return (dispatch) => {
```
tiagojsag: from here on, it's all prettier changes

frontend/.eslintrc.json
```diff
@@ -26,30 +25,14 @@
         "components": ["a"]
       }
     ],
-    "indent": [
```
tiagojsag: as per the new "eslint-check" (which is part of the online documentation of one of those eslint-prettier-config-foo modules) these rules conflict with the prettier config

sorodrigo:
Tested and looks good to go

PR #270, opened 2018-02-05
fix: moves zoom controls to top left corner

sorodrigo:
This PR includes:
- Move zoom controls to top left corner.


tiagojsag:
image

PR #275, opened 2018-02-06
feat: includes test environment and first approach to testing

sorodrigo:
Adds Jest and create a test example for the stateUrl utilities.

### DIFF
--- a/frontend/.babelrc
+++ b/frontend/.babelrc
@@ -31,7 +31,8 @@
             "components": "./scripts/components",
             "containers": "./scripts/containers",
             "constants": "./scripts/constants",
-            "router": "./scripts/router"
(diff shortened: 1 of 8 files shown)


sorodrigo:
Rebased against develop and fixed the filterStateToUrl test. I found one bug on the tool initial state of `selectedBiomeFilter`.

filtesStateToUrl was expecting ```
selectedBiomeFilter: { name: 'none' }
```
and was receiving instead ```
selectedBiomeFilter: { value: 'none' }
```

Fixed the inconsistency in the initial state and in the reducer.

I think we should merge this. LET THERE BE TESTS! 😛

agnessa:
Could you add info to the README about running these tests?

Issue #247, opened 2018-01-31
Change default basemap on map page


tiagojsag:
See https://basecamp.com/1756858/projects/12498794/todos/338651432

- [x] Change the default basemap to "default"
- [x] Change the default enabled choropleth layers to:

Brazil - soy
Production of soy

Brazil - beef
N/A

Argentina - soy
N/A

Argentina - beef
N/A

Paraguay - Soy
Production of soy

Paraguay - beef
N/A

Indonesia - palm oil

Tests:
- Different contexts have the above defined default state
- Modifying those settings on the map and reloading the page preserves the change through URL state saving

tiagojsag:
First bit done through DB settings changes and already on staging, second bit is pending merge+deploy of PR above

PR #276, opened 2018-02-06
Fix issue preveting loading default context layer

tiagojsag:
This is another side effect of setting the default state for the tool reducer. in this case, we went from undefined to an array with two nulls. so we need to take that into account

Fixes the part of  https://github.com/Vizzuality/trase/issues/247 concerning the default choropleth layer changes, together with config changes on staging (already done)


sorodrigo:
I'm wondering if `someKey: []` would be better than `someKey: [null, null]`. In any case I leave it to your discretion.

tiagojsag:
it breaks other things, I tested it

PR #289, opened 2018-02-09
Added top nodes endpoint

agnessa:
I believe both the top destination countries and top exporters visualisations can be served by this endpoint, parametrised by `context_id`, `year` and `column_id` (node type id). By default returns 10 top nodes, but can be controlled with `n_nodes` parameter.

Top exporters for context:
`/api/v3/contexts/1/top_nodes?year_start=2014&year_end=2015&column_id=6`

Top destination countries for context:
`/api/v3/contexts/1/top_nodes?year_start=2014&year_end=2015&column_id=8`

```
{
   "data":{
      "context_id":1,
      "indicator":"Trade volume",
      "unit":"Tn",
      "targetNodes":[
         {
            "id":684,
            "name":"CHINA",
            "height":0.424452686663284,
            "value":41375771.5613432
         },
         {
            "id":424,
            "name":"BRAZIL",
            "height":0.2826774040085104,
            "value":27555475.7015545
         },
         {
            "id":1840,
            "name":"NETHERLANDS",
            "height":0.0402013223456854,
            "value":3918836.61501838
         },
         {
            "id":2766,
            "name":"SPAIN",
            "height":0.02792436117120467,
            "value":2722074.86280508
         },
         {
            "id":2857,
            "name":"THAILAND",
            "height":0.02705845937460778,
            "value":2637666.50339005
         },
         {
            "id":1324,
            "name":"INDIA",
            "height":0.02207863104146035,
            "value":2152231.38658876
         },
         {
            "id":1112,
            "name":"FRANCE",
            "height":0.01710709428594789,
            "value":1667604.5351911
         },
         {
            "id":1158,
            "name":"GERMANY",
            "height":0.01624610203964705,
            "value":1583674.76016934
         },
         {
            "id":2762,
            "name":"SOUTH KOREA",
            "height":0.01611762709852366,
            "value":1571150.98547712
         },
         {
            "id":1328,
            "name":"INDONESIA",
            "height":0.01597269870893941,
            "value":1557023.32383518
         }
      ]
   }
}
```

Column ids are now returned as part of the `contexts` call:

```
         "worldMap":{  
            "mapColumnId":8,
            "listColumnId":6
         },
```

### DIFF
--- a/app/models/api/v3/context.rb
+++ b/app/models/api/v3/context.rb
@@ -19,6 +19,25 @@ module Api
       validates :country, presence: true
       validates :commodity, presence: true, uniqueness: {scope: :country}
 
+      def is_visible?
+        country.present? &&
+        commodity.present? &&
+        country_context_node_type.present? &&
+        exporter_context_node_type.present?
+      end
+
+      def country_context_node_type
+        context_node_types.find do |cnt|
+          cnt.node_type.name == NodeTypeName::COUNTRY
+        end
+      end
+
+      def exporter_context_node_type
+        context_node_types.find do |cnt|
+          cnt.node_type.name == NodeTypeName::EXPORTER
+        end
+      end
+
       def self.select_options
         Api::V3::Context.includes(:country, :commodity).all.map do |ctx|
           [[ctx.country&.name, ctx.commodity&.name].join(' / '), ctx.id]
--- a/spec/support/gold_master_urls.yml
+++ b/spec/support/gold_master_urls.yml
@@ -63,7 +63,7 @@ json:
     name: get_map_base_data
     version: v2
     v3_ready: true
-    v3_url: /api/v3/contexts/1/map_groups
+    v3_url: /api/v3/contexts/1/map_layers
     queries:
       - params: context_id=1
         name: brazil_soy
@@ -71,7 +71,7 @@ json:
     name: get_map_base_data
     version: v2
     v3_ready: true
-    v3_url: /api/v3/contexts/2/map_groups
+    v3_url: /api/v3/contexts/2/map_layers
     queries:
       - params: context_id=2
         name: paraguay_soy
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -21,6 +21,8 @@ engines:
         enabled: false
       IrresponsibleModule:
         enabled: false
+      InstanceVariableAssumption:
+        enabled: false
 
 ratings:
   paths:
--- a/app/controllers/api/v3/contexts_controller.rb
+++ b/app/controllers/api/v3/contexts_controller.rb
@@ -8,9 +8,12 @@ module Api
           includes(
             :country, :commodity, :context_property,
             readonly_recolor_by_attributes: :readonly_attribute,
-            readonly_resize_by_attributes: :readonly_attribute
+            readonly_resize_by_attributes: :readonly_attribute,
+            context_node_types: :node_type
           ).
-          all
+          references(:context_property).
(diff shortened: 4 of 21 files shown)


agnessa:
I'll add a change to `contexts` here as well, so back to WIP for a moment

agnessa:
needs to handle multi-year param

app/models/api/v3/context.rb
```diff
@@ -19,6 +19,25 @@ class Context < BlueTable
       validates :country, presence: true
       validates :commodity, presence: true, uniqueness: {scope: :country}
 
+      def is_visible?
```
tiagojsag: this seems to be slightly out of scope. could you explain how it works, please?

.rubocop_todo.yml
```diff
@@ -1,104 +1,165 @@
...
 
-# Offense count: 465
+# Offense count: 211
 # Configuration parameters: AllowHeredoc, AllowURI, URISchemes, IgnoreCopDirectives, IgnoredPatterns.
 # URISchemes: http, https
 Metrics/LineLength:
-  Max: 1095
-
-# Offense count: 44
+  Max: 80
```
tiagojsag: this is a bit more changes than I would expect. Are we addressing past issues too?

spec/support/gold_master_urls.yml
```diff
@@ -63,15 +63,15 @@ json:
     name: get_map_base_data
     version: v2
     v3_ready: true
-    v3_url: /api/v3/contexts/1/map_groups
+    v3_url: /api/v3/contexts/1/map_layers
```
tiagojsag: ?

app/models/api/v3/context.rb
agnessa: Sure, so we mentioned that each context should have an `EXPORTER` and `COUNTRY` node in order to be able to render the world map properly. We'll add checks in the validation service to detect this, but it will still be possible to have such contexts in the database. If it so happens, they should be suppressed from the `/contexts` response, and this method does a rough check for whether or not it's good to show a particular context. I guess when the validation service is merged it will be possible to replace it with a more succinct check e.g. `all_checks_passing` which checks for even more things, but in the meantime I only included this minimum to satisfy the requirements of the world map.

spec/support/gold_master_urls.yml
agnessa: This endpoint was renamed at some point, but the gold master definitions were not updated so I couldn't run the tests to check for any regressions in this PR. I still find those tests useful, so I thought better include an unrelated change here than forget about it again :)

agnessa:
Removed the rubocop config changes.

app/models/api/v3/context.rb
tiagojsag: ok, so this would be a system-wide setting, and not something limited to the world map endpoint, right? My concern here is that this might be perceived as a fully functional "hide this context in the whole site" flag, which I'm not sure we want to have/support.

spec/support/gold_master_urls.yml
tiagojsag: ah ok

app/models/api/v3/context.rb
agnessa: Maybe this is not the most clever way, but I definitely think we need a way to disable a context from being active system-wide if we have reason to expect it will break things in any of the parts of the site. For example, if it doesn't have the EXPORTER or COUNTRY node the sankey will likely break as well as the world map. I could remove this check from this PR and add a more robust version in the validation PR.

PR #292, opened 2018-02-10
Port sankey reset on no flow from old rpo

tiagojsag:


sorodrigo:
screen shot 2018-02-12 at 13 02 57

I don't really like the styles tho. Will talk with @juancarlosalonso about it.

PR #293, opened 2018-02-12
Feat/webpack perf mod

sorodrigo:
Some weekend performance mods based mainly on this [guide](https://developers.google.com/web/fundamentals/performance/webpack/).

Rebased against #275 

So let's wait until that is merged. ✌️


frontend/package.json
```diff
@@ -82,6 +91,7 @@
     "downshift": "^1.22.5",
     "find-parent": "^2.0.2",
     "highlight-words-core": "^1.1.2",
+    "jest": "^22.1.4",
```
tiagojsag: do we really need this as a prod dependency?

frontend/package.json
```diff
@@ -51,28 +53,35 @@
...
     "webpack": "^3.10.0",
+    "webpack-dashboard": "^1.1.1",
     "webpack-dev-server": "^2.9.7",
-    "webpack-merge": "^4.1.1"
+    "webpack-merge": "^4.1.1",
+    "why-did-you-update": "^0.1.0"
   },
   "dependencies": {
     "@turf/bbox": "^5.1.5",
+    "babel-jest": "^22.1.0",
```
tiagojsag: and this?

frontend/scripts/actions/tool.actions.js
```diff
@@ -420,7 +420,7 @@ export function loadLinks() {
     }
 
     const selectedBiomeFilter = state.tool.selectedBiomeFilter;
-    if (selectedBiomeFilter && selectedBiomeFilter.name && selectedBiomeFilter.value !== 'none') {
+    if (selectedBiomeFilter && selectedBiomeFilter.name && selectedBiomeFilter.name !== 'none') {
```
tiagojsag: did you test this in the browser to make sure it doesn't break something?

frontend/scripts/__tests__/utils/stateURL.test.js
```diff
@@ -0,0 +1,41 @@
+import { filterStateToURL } from '../../utils/stateURL';
```
tiagojsag: any reason for the funky folder name? can't it just be "tests"?

frontend/config/webpack.config.js
```diff
@@ -4,21 +4,39 @@ const path = require('path');
...
+  },
   output: {
     filename: '[name].[chunkhash].js',
     chunkFilename: '[name].[chunkhash].js',
     path: path.resolve(__dirname, '..', 'dist'),
     publicPath: '/'
   },
   plugins: [
-    new webpack.optimize.CommonsChunkPlugin({ name: 'common' }),
+    new webpack.optimize.CommonsChunkPlugin({ name: 'common', minChunks: 3 }),
```
tiagojsag: This doesn't seem to be test related. could you explain, please?

frontend/config/webpack.config.prod.js
```diff
@@ -6,11 +6,8 @@ const UglifyJSPlugin = require('uglifyjs-webpack-plugin');
 module.exports = merge(webpackBaseConfig, {
   devtool: 'source-map',
   plugins: [
+    new webpack.optimize.ModuleConcatenationPlugin(),
     new webpack.HashedModuleIdsPlugin(),
-    new UglifyJSPlugin({
```
tiagojsag: we're not screwing IE8 anymore? :P

sorodrigo: New version changed the config options and now support for `ie8` is false by default.

frontend/config/webpack.config.js
sorodrigo: The test specific branch is #275 . This is related to performance improvements. Sets a minimum of 3 common chunks in order to extract a module into the common bundle.

frontend/package.json
sorodrigo: 👍

frontend/package.json
sorodrigo: 👍

frontend/scripts/actions/tool.actions.js
sorodrigo: yep

sorodrigo:
@agnessa I included testing docs in the README 👍

sorodrigo:
@agnessa apparently travis is installing node by default. Adding `npm install` to `before_script` worked, tests are passing 🎉  !!!

I don't know if we should explicitly include node, what do you think?

agnessa:
great! if they’re passing l guess let’s just let them be. Specifying node explicitly might unnecessarily increase the build time; that being said, if we depend on a specific version of node and the one installed by default is not the same we might run into trouble later on, but for now it's great it was possible to get this running without big changes.

PR #295, opened 2018-02-13
feat: includes transifex script and removes polyfill io

sorodrigo:
This PR includes:
- New script for Transifex translation.
- Remove unnecessary polyfills.
- [x] Adds language selector to header.


frontend/scripts/react-components/shared/newsletter/newsletter-form.component.jsx
```diff
@@ -46,7 +47,7 @@ class NewsletterForm extends React.PureComponent {
               onInput={this.onFormInput}
               type="email"
               name="email"
-              placeholder="SIGN UP HERE TO RECEIVE UPDATES"
+              placeholder={i18n('Sign up here to receive updates').toUpperCase()}
```
tiagojsag: Isn't it better to do this with CSS rather than JS?

PR #298, opened 2018-02-14
Fix/sidekiq limit unique job execution

agnessa:
Adds more control to the sidekiq configuration and a way of ensuring long running database jobs do not execute simultaneously. Details about how that works in README.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -37,6 +37,7 @@ gem 'twitter', '~> 6.1'
 
 gem 'appsignal'
 gem 'sidekiq'
+gem 'sidekiq-unique-jobs'
 
 group :development, :test do
   gem 'byebug', platform: :mri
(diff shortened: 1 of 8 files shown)


tiagojsag:
Are we preventing a data sync (from the main db) from being scheduled if another sync process is already queued/in progress?

agnessa:
Yes, we do it on 2 levels: the protection given by this gem is the last line of defence, the first one is a custom mechanism of locking by creating a record in the `database_updates` table. To start a job you need to create a record in that table with the `status` field set to `STARTED`, and it's only possible to have one such record at any given time (which is checked both using a database partial index and an ActiveRecord validation). So theoretically it should not be possible to even add a new import job to the queue (regardless if you use the admin interface or the rake task), but just in case someone found a way to do it we have this unique sidekiq lock in place.

PR #300, opened 2018-02-15
Fix carto data loading and geojson generation

tiagojsag:
Fixes https://basecamp.com/1756858/projects/12498794/todos/335916579

Also changes the main.sh script, which was broken because we updated the topojson dependency, and for another reason I don't really know why

Test:
- Hover the map on http://localhost:8081/profile-actor?nodeId=1596&year=2015 to see the municipality names on a tooltip. In prod, we have encoding issues, which are now fixed in this branch.


sorodrigo:
Nightmare to test this in Windows. I discovered that the dev environment doesn't work in Edge nor IE. Tested in production and found that `profile-actor` page used `querySelectorAll.forEach` method, which is not supported in Windows stuff. Fixed that, tested new encodings and they're looking good!

PR #312, opened 2018-02-23
Feature/validation endpoint

agnessa:
Framework for system validations and admin interface to run them and inspect results.

Moved the description to the README: https://github.com/Vizzuality/trase/tree/feature/validation_endpoint#data-validation-process


tiagojsag:
Could we make the UI a bit more informative. Right now there's no info that the process may take a while to finish, or if the user can leave the page once the process is started.

Issue #179, opened 2018-01-04
Map: Expand bivariate maps from 9 to 16 buckets


tiagojsag:
https://basecamp.com/1756858/projects/12498794/todos/320705520
https://basecamp.com/1756858/projects/12498794/todos/322939121
https://github.com/Vizzuality/trase/issues/104

- New admin interface should allow manually changing these values
- The tool should pre-fill the yellow table bucket values with a sensible default (calculate max and min, and separate range into equal intervals).
- Some layers will have un-standard number of buckets, need to take that into account

Test
- Bi-dimension choropleth show up a 4x4 grid of values, instead of current 3x3
- Single dimension choropleth remains as-is
- Test that "0" and "na" values remain as they currently are.

agnessa:
The current bucket_5 will be an array of multiple length which represents the single dimension; it might have more than 4 values as it currently has.

The current bucket_3 is used when 2 dimensions applied and it has 2 values; it needs to be changed to allow 3 values.

tiagojsag:
on bucket_3, you may actually enforce it to have an array of 3 numbers

agnessa:
I propose to rename them `univariate_buckets` and `bivariate_buckets`

tiagojsag:
how about something simpler, like single_layer_buckets and dual_layer_buckets?

agnessa:
well I do like my Latin, but if you insist ;) sure, as long as the intention of the column is clear.

tiagojsag:
I dont insist, just seems easier to read, especially for non-natives. But pick your favorite

agnessa:
About this requirement to pre-fill with sensible values. I think we discussed that at some point not to make the default values part of the importing process for now, relying on validation endpoint + admin tool instead to sort out post-import configuration. So just double checking whether this is about setting the default value as part of post-import process (which I think is more complex than it seems, because I'm not sure how we can reliably decide whether or not e.g. a new map attribute should be created), or perhaps when someone goes to the admin tool, starts adding a new map attribute and the buckets come up populated in the form?

Another question, how do I rescale the current values for bucket_3?

PR #314, opened 2018-02-26
Feature/expand map buckets

agnessa:
- renamed `bucket_3` -> `dual_layer_buckets`, `bucket_5` -> `single_layer_buckets`
- `dual_layer_buckets` needs to be 3 elements (values rescaled by adding an element in between)
- `single_layer_buckets` can be variable length; because of that there is now an sql function which calculates the bucket for a given value in the `node_attributes` endpoint (we used to have a `CASE` statement there, but that wouldn't work for variable length arrays)
- while making the changes to the `map_layers` controller I noticed it still uses v2 serializers, so I added v3 versions so that it's easier to get rid of v2 in the future
- did not add the capability to populate buckets automatically

`bundle exec rake db:migrate`


agnessa:
https://github.com/Vizzuality/trase/issues/179

Did not add the capability to populate buckets automatically, as I'm not sure at which point this is supposed to happen (import script? admin tool?) and also I think there are more properties that could benefit from default values, and not sure if providing those is in scope.

I believe all of these properties could be prepopulated from data in `flows`, and to me it would make sense to either do it for all of them or none:
    - `map_attributes#years`
    - `map_attributes#single_layer_buckets`
    - `map_attributes#double_layer_buckets`
    - `download_attributes#years`
    - `ind/qual/quant properties#is_temporal`
    - `recolor_by_attributes#years`
    - `recolor_by_attributes#min/max_value`
    - `resize_by_attributes#years`

agnessa:
@sorodrigo I believe this is ready for you to look at. The first thing to look into is handling the 4x4 grid for dual layers. Not sure what needs doing with single layers at this point (eventually they need to support variable length buckets.)

tiagojsag:
Backend changes look good, but can't be merged until we do the FE changes, due to the increase in number of buckets

tiagojsag:
This will need another iteration once https://github.com/Vizzuality/trase/issues/323 is done, but for now we should merge, to avoid rebase nightmare (and to avoid me having to rollback my database, im lazy).

PR #319, opened 2018-02-28
Feat/Responsive profile pages

tsubik:
Responsive profile pages:

- Profile place page (https://github.com/Vizzuality/trase/issues/306)
- Profile actor page (https://github.com/Vizzuality/trase/issues/307)
- Profile search page (https://github.com/Vizzuality/trase/issues/305)

Switching to tooltip.js powered by popper.js to fix issues with tooltips on small resolutions.


tsubik:
Making it "WIP" because profile actor page is very similar and maybe it would be easier to review changes to make both those pages responsive.

tsubik:
I hope that's it ;).

tiagojsag:
Tested on iPhone X, iPhone 6S+ and iPad

Search page
- [ ] Placeholder text seems very small. Let's review with @juancarlosalonso 
- [ ] Loading spinner is cut off
- [ ] On iPhone, clicking on the input zooms the page in, which in itself is not a big deal. However, if you select an actual actor/place from there, the page does not zoom out, which makes the actor/place profile page look broken.
- [ ] Tooltips are not working on mobile - only the scatterplot ones are.
- [ ] The new selectors for the tabbed tables seem to have the wrong fonts.

tsubik:
I will fix tooltips problem in a few hours when I'm back. Looks like this is a known problem in popper.js library, it only does not work on iOS.

tsubik:
Should be fine now :), tested on iPad.

PR #321, opened 2018-03-01
Added years of data avalable for recolor / resize

agnessa:
Include years available for recolor / resize in contexts endpoint


agnessa:
I believe serving the list of years is all that's required backend-wise to support greying out resize / recolor by when data not available?

PR #326, opened 2018-03-02
Feat: explore page

sorodrigo:
This PR includes the new explore page that will appear before the sankey:
- World Map
- Top Exporting Companies list
- [x] conditionally redirect to `/explore`
- [x] add link to `/flows` and to `/terms-of-use`
- [x] change top exporters values to % in `api`

world-map


sorodrigo:
SuperReady! Sorry for the huge list of commits. I had to merge instead of rebase because Git went loco. @tiagojsag

tiagojsag:
- [ ] Styling is not consistent on 1st load vs going "back" from explore
- [ ] Countries have links to profile pages
- [ ] Labels on the map flows are quite small and hard to read.

PR #353, opened 2018-03-13
Prevent before_save callback from setting highlighted to false

agnessa:
this happened when you edited a highlighted post and did not change the value of the highlighted flag - it would de-highlight itself because of the `before_save` callback

### DIFF
--- a/app/models/content/post.rb
+++ b/app/models/content/post.rb
@@ -34,7 +34,8 @@ module Content
     validates_attachment_content_type :image, content_type: /\Aimage\/.*\z/
 
     after_initialize :set_default_date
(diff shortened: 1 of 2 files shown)


app/models/content/post.rb
```diff
@@ -46,7 +46,7 @@ def set_default_date
     end
 
     def reset_highlighted_flag
-      return true unless highlighted
+      return true unless highlighted && highlighted_changed?
```
tsubik: or maybe we should just conditionally invoke that callback
```
before_save :reset_highlighted_flag, if: -> { highlighted && highlighted_changed? }
```
what do you think @agnessa ?

PR #359, opened 2018-03-13
Fix/remove old tables

agnessa:
To be merged after https://github.com/Vizzuality/trase/pull/352

- Removed all tables (except AR ones), materialized views, user defined functions and data types from the `public` schema
- Moved all tables, mviews & functions from `revamp` to `public`
- Removed the logic that used to handle separate database connections
- Removed any leftover namespacing
- Removed prosthetic delete & commit statments in specs, which are no longer necessary since database cleaning works again using transaction rollback
- Updated schema annotations
- **PLEASE NOTE CHANGE IN `.env`**:
  `TRASE_LOCAL_SCHEMA=public` - it affects the `postgres_fdw` setup and the import script


tiagojsag:
@agnessa LGTM, just needs to be reviewed to fix conflicts

agnessa:
conflicts resolved

PR #360, opened 2018-03-13
Disable database updates & database validations in production

agnessa:
Just so that no-one is tempted, this hides the relevant sections of the admin menu in production. Those operations are only meant to be run in integration environment.


tiagojsag:
@agnessa could you re-check this, I am still seing the menus with RAILS_ENV=production

PR #365, opened 2018-03-14
Fix/site dives

tsubik:
Restoring site dives function.

screenshot from 2018-03-14 12-41-33


tsubik:
ahh wait, there is one new console warning I'd like to fix.

tsubik:
Ok. Done

PR #366, opened 2018-03-14
Cache invalidation

agnessa:
Expected behaviour: "when I'm using the admin tool to make changes to content on the website, I can see the effect of my changes immediately". That is currently not working because of cache configuration.

This PR adds Varnish cache invalidation using PURGE and BAN requests. The first type is used to clear exact urls, the second one to clear all urls matching a regexp. Purging has been added to all relevant admin controllers.

Also, cache headers are now only set in production, so that in staging it's easy to reload values from the API.

**Important**: required changes to Varnish configuration are described in README. Those have been already applied on staging, need to be yet applied on production: https://github.com/Vizzuality/trase/issues/372

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -25,6 +25,10 @@ Layout/MultilineOperationIndentation:
 Layout/SpaceInsideHashLiteralBraces:
   EnforcedStyle: no_space
 
+Lint/RescueWithoutErrorClass:
+  Exclude:
+    - 'app/controllers/concerns/cache_utils.rb'
+
 Metrics/BlockLength:
   Exclude:
     - 'spec/**/*'
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -33,6 +33,7 @@ Metrics/BlockLength:
   Exclude:
     - 'spec/**/*'
     - 'lib/tasks/**/*'
+    - 'app/admin/**/*'
 
 Metrics/LineLength:
   Max: 80
--- a/README.md
+++ b/README.md
@@ -64,6 +64,55 @@ To deploy, simply use:
 cap <staging|production> deploy
 ```
 
+We use Varnish in staging / production for caching. The following configuration needs to be present to enable cache purging in `/etc/varnish/default.vcl`:
+
+```
+acl purge {
+    "localhost";
+}
+
+sub vcl_recv {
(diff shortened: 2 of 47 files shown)


agnessa:
To achieve the expected result I propose 3 changes, currently applied only to `/content/posts`

- purge relevant varnish cache after every update operation
- use `fresh_when` method in controller to set the `last-modified` header
- use the `no-cache` cache mode in the `fetch` function to force conditional queries with `If-Modified-Since`

@tiagojsag @tsubik ?

tsubik:
Hmm, I think that the server should control how the resource is cached, and as this is a dynamic resource we probably should not cache it on the browser and have Cache-Control set to something like `max-age=0, private, must-revalidate`.

tiagojsag:
I'm not super concerned about client side cache being invalidated. Content editors know they are supposed to clear their browser cache after they make updates, the pain point right now is that we have a cache (varnish) that they can't clear. I'm happy with just invalidating varnish

tsubik:
Right, as I see we are using `max-age=7200` so that's the maximum time of the content to be stale. After a second thought, I think that's ok for the users and I agree with you @tiagojsag that varnish invalidate should be enough.

agnessa:
@tiagojsag - how would the editor refresh the content if ctrl+r does not work? We need to advise them.

tiagojsag:
@agnessa pressing ctrl+r again :P at some point, the browser will bypass its own cache. I don't think the UI needs to inform the users, but if there's a low hanging fruit on this, lets go for it.

agnessa:
@tsubik would that make the website reload the resource every time within the same session? I'm thinking of using the same approach for caching content, such as posts, but also all the other responses from the API. We now allow to make changes to website configuration using the admin tool, and that is the point at which we need to be able to make a change and observe the effect, this would be mostly done in the integration environment though; in production I suppose we might not want to reload API responses this way. Do we need a different approach for staging / production?

agnessa:
@tiagojsag I believe that point will occur after 2 hours, as that is our current expiration time :) I haven't been able to refresh content this way under a tighter time constraint :)

tsubik:
hmm, it should work, maybe `must-revalidate` must be present too in the headers? And actually that `2 hours` of stale data, for news that might be too long ;)

PR #368, opened 2018-03-15
Prevent dual state change on homepage map

tiagojsag:
This feels slightly evil, so let's review properly and consider alternatives

### DIFF
--- a/frontend/scripts/react-components/shared/world-map/world-map.component.jsx
+++ b/frontend/scripts/react-components/shared/world-map/world-map.component.jsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import {
   ComposableMap,
@@ -14,7 +14,7 @@ import UnitsTooltip from 'react-components/shared/units-tooltip.component';
 import cx from 'classnames';
 import formatValue from 'utils/formatValue';
 
-class WorldMap extends React.PureComponent {
+class WorldMap extends Component {
   static buildCurves(start, end, arc) {
     const x0 = start[0];
     const x1 = end[0];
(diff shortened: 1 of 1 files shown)


tsubik:
I think we can add some parameter `countriesAnnotations` with default value `true` and just not render those annotations on home page at all.

tsubik:
Ahh forget it, the same problem I see is on explore page.

tsubik:
Maybe rendering annotations if the context is not selected?

```
<Annotations>{flows.length === 0 && !selectedContext && this.renderCountriesAnnotations()}</Annotations>
```

tiagojsag:
I'm not sure that would fix the blinking problem of the flows themselves

tsubik:
Right, it won't.

frontend/scripts/react-components/shared/world-map/world-map.component.jsx
```diff
@@ -44,13 +44,18 @@ class WorldMap extends React.PureComponent {
     this.renderCountriesAnnotations = this.renderCountriesAnnotations.bind(this);
   }
 
-  componentDidUpdate(prevProps) {
+  shouldComponentUpdate(nextProps) {
```
tsubik: I think `shouldComponentUpdate` should not have side effects like loading the data. Without introducing extra property into the state like `isLoading` we could avoid this flickering in a similar way.

```
  componentWillReceiveProps(nextProps) {
    if (
      nextProps.selectedContext !== this.props.selectedContext ||
      nextProps.selectedYears !== this.props.selectedYears
    ) {
      this.props.getTopNodes();
    }
  }

  shouldComponentUpdate(nextProps) {
    // data is probably being loaded, do not update component yet
    if (nextProps.flows.length === 0 && this.props.flows.length > 0) {
      return false;
    }

    return true;
  }

```

sorodrigo: I agree side-effects on `shouldComponentUpdate` is a no go.

PR #369, opened 2018-03-15
Fix/ios safari video issues

tsubik:
Fixing issues with video player mostly for iOS

- the player will no longer restart itself after the end, it should just close itself,
- fixing fullscreen iPad z-index issues with feedback button and navbar,
- fixing player not showing up on iPhone at all,
- improving interaction with movie progress slider on iOS using https://rangetouch.com/. Looks like only sliding worked and clicking/touching the track didn't.

### DIFF
--- a/frontend/scripts/index.ejs
+++ b/frontend/scripts/index.ejs
@@ -9,6 +9,4 @@
 <div id="app-root-container"></div>
 </body>
 <script type="text/javascript" src="//cdn.transifex.com/live.js"></script>
-<!-- Rangetouch to fix <input type="range"> on touch devices (see https://rangetouch.com) -->
-<script src="https://cdn.rangetouch.com/1.0.5/rangetouch.js"></script>
 </html>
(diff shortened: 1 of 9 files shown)


frontend/scripts/index.ejs
```diff
@@ -9,4 +9,6 @@
 <div id="app-root-container"></div>
 </body>
 <script type="text/javascript" src="//cdn.transifex.com/live.js"></script>
+<!-- Rangetouch to fix <input type="range"> on touch devices (see https://rangetouch.com) -->
+<script src="https://cdn.rangetouch.com/1.0.5/rangetouch.js"></script>
```
tiagojsag: cant we use https://github.com/sampotts/rangetouch#node--npm instead?

tsubik: Done.

tsubik:
@tiagojsag I think that we should allow users to pause video without closing it. To close video the user will have to leave the fullscreen mode or it will close automatically when it ends. Any thoughts?

tsubik:
Ready :rocket:

tiagojsag:
Let's wait for rodrigo to review this

tsubik:
Done :)

PR #370, opened 2018-03-15
place tabs for scatterplot at the bottom

tsubik:
Placing tabs for scatterplot at the bottom of the chart as it was before.

image


tsubik:
Deforestation table header alignment fixed here too.

Issue #202, opened 2018-01-11
Limit "recolor by" and "resize by" by year


tiagojsag:
SANKEY: Automated solution for visualizing / greying out indicators (resize or recolour) when they are not available for specific years

PR #373, opened 2018-03-16
Grey out recolor by and resize by options when not available for sele…

tiagojsag:
…cted year

Finalizes https://github.com/Vizzuality/trase/issues/202

Also adds a missing option on a hardcoded list on the API


sorodrigo:
Please add more thorough testing info. I don't know when this is not available.

tiagojsag:
This is configurable on the admin UI. For each recolor by/resize by criteria, you can specify which years are available, and which are not. That information should then be translated into greyed out options on the frontend

PR #377, opened 2018-03-19
fix: avoids race condition and removes event listeners

sorodrigo:
This PR fixes crash when going to profile pages through the sankey and then clicking back button. Also removes event listeners for sanity's sake.


tiagojsag:
let's chat about this

tsubik:
Would be great to perform proper cleanup using that pattern for all not react components.

Issue #382, opened 2018-03-21
When navigating between different profile pages, the page should scroll up automatically


tiagojsag:
Steps to reproduce:
- Open a profile page
- Scroll down
- Click on a link to a different profile page

The new page is loaded at the bottom, not at the top

Issue #333, opened 2018-03-07
Language not selected on Safari


tsubik:
When I clear all the app related data from localStorage and sessionStorage and reopen any page without having language set in URL then the language seems to be not selected. I noticed that on Safari, and all devices I've checked on have only Polish language set. 

![image](https://user-images.githubusercontent.com/1286444/37087551-e652bef2-21fa-11e8-95f5-15590026f3d5.png)

tiagojsag:
I believe this is fixed, closing

tsubik:
Just checked and it's unfortunately not, at least on my iPad that problem is still present. Tested also on Chrome on that iPad and the same issue.

PR #384, opened 2018-03-21
fix: restores scroll on page change

sorodrigo:
Restores page's scroll to top on navigation.

Addresses #382

### DIFF
--- a/frontend/config/webpack.config.js
+++ b/frontend/config/webpack.config.js
@@ -17,7 +17,7 @@ const templates = require('./static.templates');
 
 module.exports = {
   entry: {
-    main: path.join(srcPath, 'index')
+    main: ['babel-polyfill', path.join(srcPath, 'index')]
   },
   output: {
     filename: '[name].[chunkhash].js',
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -3051,71 +3051,11 @@
         "request": "2.85.0"
       },
       "dependencies": {
-        "caseless": {
-          "version": "0.12.0",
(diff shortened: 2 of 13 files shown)


tiagojsag:
Seems to only work on the first try, after that it fails.

sorodrigo:
@tiagojsag I fixed all our issues regarding the scroll restoration.
@tsubik I think I solved the safari issue #333 

Please review :)

tsubik:
Hmm, I still have this language selector issue :/ @sorodrigo 
About the scroll restoration, surely we should restore to the top when navigating to a new page. Although, as a user, when clicking back button (especially on mobile), I would expect to see the exact same spot where I was, not the top of the page.

sorodrigo:
> I would expect to see the exact same spot where I was, not the top of the page.

That's easy to achieve, thoughts about this @tiagojsag ?

tiagojsag:
not super concerned about restoring scroll when pressing back, to be honest. If it's easy to do (a couple of hours) sure, let's go for it, but otherwise I'm happy scrolling to the top

tiagojsag:
when I use back, I am not taken to the top of the page, but rather stay in the same scroll position as i was when I hit the back button. This feels weird, I think perhaps a better solution would be to scroll to the top too?

tsubik:
Right, it feels weird, but using that restore-scroll I was able to make `about` page works just fine, going back and forward to left scroll positions, but when opening the new page going to the top. Unfortunately, not profile pages :/. What I also noticed there is always the second redirect caused by locale selector when changing page, probably unnecessarily.

Issue #191, opened 2018-01-09
Improve further the profile pages print to pdf mode


sorodrigo:
See https://basecamp.com/1756858/projects/12498794/todos/327056250

- [x] The chord diagrams won't fit in an A4 page with their current dimension, we need to shrink them when print query-param is true. (Looks like this is obsolete)
- [x] The choropleth legends colors arent showing up (just like background isn't either).
- [x] The map outline should be white when print query-param is true.
- [x] Keep table/chart title on the same page
- [x] Remove tooltips
- [x] Scatterplot chart should be repeated for all parameters (Land use, Territorial def...)

tsubik:
So the background color and other graphics problem is the browser settings thing.
![image](https://user-images.githubusercontent.com/1286444/37603111-2427aa48-2b8e-11e8-8c02-917c9576ac25.png)
The question is can we set this setting for that webshot API? @sorodrigo @tiagojsag

sorodrigo:
So apparently there's an option called `printBackground` https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#pagepdfoptions

tiagojsag:
Here's how the pdfs currently look:

[TRASE-1523281593872.pdf](https://github.com/Vizzuality/trase/files/1890141/TRASE-1523281593872.pdf)
[TRASE-1523281720245.pdf](https://github.com/Vizzuality/trase/files/1890143/TRASE-1523281720245.pdf)

tiagojsag:
@juancarlosalonso what do you think? The only thing I'm noticing is we don't have the trase brand anywhere, and we probably should.

sorodrigo:
> The map outline should be white when print query-param is true.

I think this is missing.

juancarlosalonso:
Apart from the map outline (that should be white) and adding a the name of the company as a label close to the scatterplot red points, looks good to me

tsubik:
Here are some examples with white outlines. I removed them completely for small maps. I also brought red dot to front, but I'm not sure about that idea of labeling that dot.

[TRASE-1523369771420.pdf](https://github.com/Vizzuality/trase/files/1894998/TRASE-1523369771420.pdf)
[TRASE-1523369308027.pdf](https://github.com/Vizzuality/trase/files/1894999/TRASE-1523369308027.pdf)

tiagojsag:
not sure if close or not @tsubik @juancarlosalonso

tsubik:
[...]

PR #386, opened 2018-03-21
Feature/Improving profile pages print mode

tsubik:
List of some of the improvements copied from #191 

- [x] The choropleth legends colors arent showing up (just like background isn't either).
- [x] The map outline should be white when print query-param is true.
- [x] Keep table/chart title on the same page
- [x] Remove tooltips
- [x] Scatterplot chart should be repeated for all parameters (Land use, Territorial def...)

Best way to test locally: add `print=true` param to query string and just print in your browser. I've tried with local webshot API, but even using `waitFor` certain elements I always had an issue with something not rendering (like maps). I haven't seen those issues on staging or prod.

### DIFF
--- a/frontend/html/profile-actor.ejs
+++ b/frontend/html/profile-actor.ejs
@@ -126,7 +126,7 @@
           </div>
         </section>
 
-        <section class="c-scatterplot-container page-break-inside-avoid">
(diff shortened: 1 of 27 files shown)


tsubik:
Done @tiagojsag

PR #387, opened 2018-03-22
Feature/improve dependencies

tiagojsag:


frontend/package.json
```diff
@@ -45,80 +47,76 @@
...
     "style-loader": "^0.19.1",
-    "uglifyjs-webpack-plugin": "^1.2.2",
     "url-loader": "^0.6.2",
-    "webpack": "^3.11.0",
-    "webpack-dashboard": "^1.1.1",
-    "webpack-dev-server": "^2.11.1",
+    "webpack": "^4.2.0",
+    "webpack-cli": "^2.0.12",
+    "webpack-bundle-analyzer": "^2.11.1",
+    "webpack-bundle-size-analyzer": "^2.7.0",
```
sorodrigo: whats the difference between `webpack-bundle-analyzer` and `webpack-bundle-size-analyzer`

tiagojsag: one we need, the other I will remove, because I think it's not needed :P

frontend/scripts/index.js
```diff
@@ -25,6 +24,9 @@ if (process.env.NODE_ENV !== 'production' && PERF_TEST) {
 }
 
 if (process.env.NODE_ENV !== 'production' && REDUX_LOGGER_ENABLED) {
+  // eslint-disable-next-line global-require,import/no-extraneous-dependencies
+  const { createLogger } = require('redux-logger');
```
sorodrigo: I like this, but then lets use also global require for React in the line 21 of the same file.

frontend/config/webpack.config.js
```diff
@@ -4,7 +4,7 @@ const path = require('path');
 const webpack = require('webpack');
 const HtmlWebpackPlugin = require('html-webpack-plugin');
 const CleanWebpackPlugin = require('clean-webpack-plugin');
-const InlineChunkWebpackPlugin = require('html-webpack-inline-chunk-plugin');
+// const BundleAnalyzerPlugin = require('webpack-bundle-analyzer').BundleAnalyzerPlugin;
```
sorodrigo: should we remove this?

tiagojsag: remove no, document yes

frontend/package.json
```diff
@@ -35,8 +36,9 @@
     "autoprefixer": "^7.2.6",
     "babel-core": "^6.26.0",
     "babel-eslint": "^8.2.2",
-    "babel-jest": "^22.4.1",
-    "babel-loader": "^7.1.3",
+    "babel-jest": "^22.4.3",
+    "babel-loader": "^7.1.4",
+    "babel-plugin-lodash": "^3.3.2",
```
tsubik: Do we actually need that? Wouldn't be better to replace all `import _ from 'lodash'` with just import the function that we need?

frontend/scripts/index.js
tsubik: :+1: I agree

frontend/package.json
tiagojsag: I prefer this for two reasons:
- this is equivalent, and allows us to not have to worry about that
- this will most likely be more reliable at that task than us

less busy and lazier developers is what the world needs ;)

sorodrigo: Fine, but let's keep enforcing individual imports

tsubik: Hehe ok, I just always try to minimalize the number of dependencies.

tiagojsag: indeed, but I am more concerned about dependencies that make it to the user's browser, and not so much with serverside stuff

PR #391, opened 2018-03-22
Add prefetching to all webpack chunks

tiagojsag:


sorodrigo:
I'm not sure if I want to see it faster, but damn it looks a lot faster with prefetch + webpack 4!!! 🔥 🔥 🔥

PR #392, opened 2018-03-22
Feature/scroll and transifex dont play nice

tiagojsag:
Tested that the routerListener is not called too often.
Still need to test URL query params update (link between two pages of same type)

### DIFF
--- a/frontend/scripts/router/route-subscriber.js
+++ b/frontend/scripts/router/route-subscriber.js
@@ -1,4 +1,5 @@
 import connect from 'connect';
+import { updateScroll } from 'redux-first-router';
 
 export default function routeSubscriber(store) {
   class RouterComponent {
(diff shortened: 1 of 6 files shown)


frontend/scripts/router/route-subscriber.js
```diff
@@ -20,6 +21,7 @@ export default function routeSubscriber(store) {
 
     onRouteChange({ routesMap, type } = {}) {
       const filename = routesMap[type].page;
+      updateScroll();
```
sorodrigo: I don't like this. If the query params change and the page didn't it will scroll to top.

tiagojsag: I thought so too, but no. Tested on the sankey, but will retest again tomorrow.

PR #394, opened 2018-03-23
Add delayed service worker to pre-cache webpack chunks

tiagojsag:


tiagojsag:
@sorodrigo is affraid of it, so maybe we can add a feature flag to test on staging for a while, before going to prod with it :fearful:

sorodrigo:
I'm just gonna move the service worker code into the JS files, and merge. We can't really add a flag, so let's release it to staging and test away!

tsubik:
Maybe we could try to `unregister` service worker if the flag is not set?

sorodrigo:
We could I just don't see why, maybe I'm not understanding the reason why we need the flag? Are you trying to add/remove this feature with an env var so that we avoid having to remove the code if we finally decide against it? or why? @tiagojsag

tiagojsag:
my idea is that we only load the service worker if the flag is true. the unregister idea seems nice, yes

sorodrigo:
So I replaced the regiter in the `index.ejs` for a `worker.js` file. This is ~a rip off~ inspired in the implementation of `create-react-app` and considers a lot of edge cases and add the possibility to add callback hooks.
Also, I tested the timeout and didn't make any difference, this is mainly because a service worker intercepts requests to cache, but doesn't really affect the order in which they are made. Please review, I think it's ready.

PR #396, opened 2018-03-23
Remove unnecessary static files

tiagojsag:


sorodrigo:
I dunno, do we need to get rid of the dev mocks?

tiagojsag:
For me they make sense to mock an API we don't have, because it's not implemented, unstable, or something else. Since that's no longer the case, I don't see a reason to keep them around

PR #398, opened 2018-03-23
Feature/tool to copy databases

agnessa:
Description in README - requires changes to `.env` file (templates in basecamp updated).


tiagojsag:
As we chatted over slack
- We should use the no-owner flag to avoid conflict on importing
- We should store a copy of the current database before importing, and fallback to it in case the import fails for any reason. If attempt once to restore that backup, and if that fails too, we call a ~adult~ human
- would be good to have a CLI command to restore a database dump, to use in the scenario above

agnessa:
Now with force option available, which ignores the schema compatibility check - for when you know what you're doing.
`DATABASE_VERSION="AF DEV/20180403-12:06:25+02:00.dump.gz" bundle exec rake db:s3:force_import`

PR #403, opened 2018-03-27
Download chunking

agnessa:
A few changes here:
- some new indexes on `download_flows_mv` to speed up filter by context and ordering of rows in downloads
- new column `row_name` in the `download_flows_mv`; this one used to be calculated on the fly for purposes of the pivot download, so having it pre-calculated should give it a boost and also can be used for sorting rows
- sorting is important, because we now added the chunking functionality to split files longer than 500k rows; this is done using `LIMIT` and `OFFSET` and it requires a predictable sort order
- my favourite part is giving the json downloads a boost by replacing the Rails `to_json` method with native Postgres json support. It makes the query take longer but there's almost no overhead on top of the query and the total time improvement is remarkable (real time):

```
require 'benchmark'

def rails_to_json(collection_query)
  collection_query.to_json
end

def postgres_to_json(collection_query)
  json_query = Api::V3::Readonly::DownloadFlow.
    select('array_to_json(array_agg(row_to_json(t)))').
    from("(#{collection_query.to_sql}) t")
  result = Api::V3::Readonly::DownloadFlow.connection.
    execute(json_query.to_sql)
end

collection_query = Api::V3::Readonly::DownloadFlow.where(context_id: 1).limit(500000)

collection_query.size

Benchmark.bm do |x|
  x.report("Postgres to_json:") { postgres_to_json(collection_query) }
  x.report("Rails to_json:") { rails_to_json(collection_query) }
end
                          user     system      total        real
Postgres to_json:     0.390000   0.650000   1.040000 ( 14.345017)
Rails to_json:      159.350000  13.880000 173.230000 (185.115569)
```

### DIFF
--- a/app/controllers/api/v3/download_controller.rb
+++ b/app/controllers/api/v3/download_controller.rb
@@ -6,13 +6,13 @@ module Api
 
         respond_to do |format|
           format.csv do
-            send_file download.zipped_csv.path,
+            send_data download.zipped_csv,
                       type: 'application/zip',
                       filename: "#{download.download_name}.zip",
                       disposition: 'attachment'
           end
           format.json do
-            send_file download.zipped_json.path,
+            send_data download.zipped_json,
                       type: 'application/zip',
                       filename: "#{download.download_name}.zip",
                       disposition: 'attachment'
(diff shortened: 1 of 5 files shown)


agnessa:
Implement chunking by year

PR #404, opened 2018-03-27
Feature/responsive explore page

tsubik:
- Making the explore page more responsive, adding the mobile menu for small resolutions
- Showing the world map with the link to the explore page on the home page for all resolutions
- Fixing a year selector so changing the year updates flows on the world map
- Fixing proper context load for the world map
- Adding support for touch events for a year selector

explore-page-responsive

Home page

responsive-homepage-worldmap

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -3842,7 +3842,7 @@
     },
     "compression": {
       "version": "1.7.2",
-      "resolved": "http://registry.npmjs.org/compression/-/compression-1.7.2.tgz",
+      "resolved": "https://registry.npmjs.org/compression/-/compression-1.7.2.tgz",
       "integrity": "sha1-qv+81qr4VLROuygDU9WtFlH1mmk=",
       "dev": true,
       "requires": {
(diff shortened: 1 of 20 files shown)


tsubik:
I'm going to replace the current year selector with a normal dropdown.

sorodrigo:
@tsubik we can't do that, because in the near future we will re-enable multiple year selection.

tiagojsag:
- [ ] "SELECT BOTH A COUNTRY AND A COMMODITY" font is incorrect
- [ ] Popups show up offcanvas (test with china for indonesian palm oil)
- [ ] I'm wondering if we should put a message like "if you want to explore the full supply chain, visit trase.earth on a computer" or sorts. My concern is that first time users may think this is it.

tiagojsag:
Almost there, just a couple of thing missing to be perfect:
- on an actual phone, there's a weird effect if you click on any country on the map.
- seems like the tooltip only shows up if you click on the flow itself, which is really hard to do on a phone.

tsubik:
Both should be fixed now @tiagojsag

PR #406, opened 2018-03-28
Add support for canvas-based rendering of choropleth data

tiagojsag:
Adds support for canvas-based rendering for the choropeth layer. Still supports svg-based rendering on browsers that don't support canvas, or through a config flag

The whole SVG rendering part now relies on per-element styling, rather than css-based classes. Since we reuse some of those classes elsewhere, we can't delete them all, but we could definitely remove a bunch of them.

Key stuff to test here is performance. We should only push this if there's something to gain, otherwise may not be worth


tsubik:
I've tried to experiment with that, and on my machine maybe I can see a slight performance improvement. It's hard to test it without blowing up dev tools :P. Maybe there is a way to increase the performance by using many smaller canvas tiles instead of one. There is https://github.com/Leaflet/Leaflet.VectorGrid but honestly, I haven't played around with it yet, and I'm not sure if we can do everything we need with it.

PR #408, opened 2018-03-28
Rebased/data portal conditional queries

tiagojsag:


tiagojsag:
- [ ] @agnessa API is now returning ugly names instead of human-friendly names for indicators

tsubik:
I will handle that `display_name` @agnessa. I'm feeling dumb that I have dismissed that issue as a "bad data" one as I recently restored local DB from production dump.

tsubik:
Done @tiagojsag

PR #410, opened 2018-03-28
Feature/Not supported on mobile page

tsubik:
Pages like tool, map and data download are not supported on mobile screens. The application will redirect to the new "not supported" page if anyone will try to open any of those pages.

image


tiagojsag:
great stuff!

PR #413, opened 2018-03-29
Feature/varnish cache warming

agnessa:
I propose to go easy with this, using a periodic task run by cron to warm it. I did not include logic to warm it after purging, because I guess we can work around by warming more often if that becomes problematic. This PR includes:
- a rake task to regenerate the public/urls.txt file (actually the file contains paths rather than urls to allow using it in all environments):
    `bundle exec rake cache:warmer:generate_urls`
     At this point this is top 10 municipalities in Brazil and top 10 exporters; there is a quiet assumption here, unverified by any visits analysis, that "bigger" = "more popular" as well as less controversial "bigger" = "longer to generate"
- a rake task to warm the cache by hitting said urls (requires `API_HOST` var to be set):
` bundle exec rake cache:warmer:run`
- a `whenever` schedule with Capistrano integration which will update the crontab on deploy to run the cache warming rake task every 2 hours


tiagojsag:
I agree with the approach :+1: 

One thing: keep in mind that each profile page has multiple years, which need to be cached individually.

agnessa:
Need to warm cache per year

agnessa:
Added caching for available years

tiagojsag:
feels like the URL file should be ignored by git, because it's DB dependent, not code dependent. 
Also: we should probably auto-refresh it on db import

agnessa:
makes sense on data import, but also probably for data upload from main db

tiagojsag:
True, I knew I was forgetting something ;)

agnessa:
removed `urls.txt` from source control and added logic to regenerate it after data update / import; also, there are now rake tasks to clear cache & regenerate the cache warming urls

PR #418, opened 2018-04-05
Feature/team page a few changes for smaller resolutions

tsubik:
This PR introduces team profiles carousel for smaller resolutions as per design, and also a couple of small style changes. 

responsive-team-page

### DIFF
--- a/frontend/scripts/react-components/team/team.component.jsx
+++ b/frontend/scripts/react-components/team/team.component.jsx
@@ -25,7 +25,7 @@ class Team extends PureComponent {
               })}
               style={{ backgroundImage: `url(${members[slug].smallImageUrl})` }}
             />
-            <h3 className="team-list-item-title title -medium">{members[slug].name}</h3>
+            <h3 className="team-list-item-title title -medium -light">{members[slug].name}</h3>
             <span className="team-list-item-subtitle subtitle -gray">See More</span>
           </Link>
         </div>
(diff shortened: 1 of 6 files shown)


tiagojsag:
Looks good to me, what do you think @sorodrigo ?

frontend/scripts/react-components/team/team.component.jsx
```diff
@@ -19,24 +25,46 @@ class Team extends PureComponent {
...
 
     return (
       <section className="team-group" key={group.name}>
         <h3 className="subtitle">{group.name}</h3>
         <div className="team-list">
-          <div className="row -equal-height">
-            {group.staffMembers.map(slug => this.renderTeamMember(slug))}
-          </div>
+          <ResizeListener>
+            {({ resolution, windowWidth }) =>
```
sorodrigo: pretty neat!

frontend/scripts/router/nav-links.js
```diff
@@ -50,10 +50,10 @@ export default {
         payload: { section: 'faq' }
       }
     },
-    // {
-    //   name: 'Team',
-    //   page: 'team'
-    // },
```
sorodrigo: Are we sure about this?

frontend/styles/components/static-content/team.scss
```diff
@@ -10,6 +10,11 @@
 
   .team-list {
     padding: 20px 0 0;
+
+    @media screen and (max-width: $breakpoint-foundation-small) {
```
sorodrigo: I prefer min-width, I know we've talked about this before and the project has a mix of both so not super important.

frontend/scripts/router/nav-links.js
tiagojsag: I'm sure we need to revert this before merging ;)

tsubik: Hmm, dunno when that should be enabled, maybe you are right that not today :D, will comment this out again.

sorodrigo:
OK looks great, I have just one question. When entering the team-member page should the `SidebarNav` display team as active? This is not related to your changes but I realised that it doesn't.

tsubik:
Fixed @sorodrigo

Issue #221, opened 2018-01-18
Inspect Autoprefixer in production


sorodrigo:
It seems the autoprefixer is not doing its job. 😒

PR #428, opened 2018-04-06
fix: enable autoprefixer for scss files

tsubik:
Autoprefixer was not included in the pipeline for `.scss` files.


sorodrigo:
I have to test this, but if it works you have no idea how much time I wasted trying to fix it and couldn't. 😂 

Thanks a lot!

#221

tsubik:
Ohh, I haven't seen that issue ;). I was poking around with my newest iOS sankey issue and found out that there are no prefixes ;P

tiagojsag:
I have no clue what this is or does, but it smells like CSS, so I am running away. have fun with it @sorodrigo

Issue #199, opened 2018-01-11
Expand new search to data portal


tiagojsag:
See #108

Issue #198, opened 2018-01-11
Expand new search to static pages


tiagojsag:
See https://github.com/Vizzuality/trase/issues/108

Issue #197, opened 2018-01-11
Expand new search to profile pages


tiagojsag:
See https://github.com/Vizzuality/trase/issues/108

PR #448, opened 2018-04-13
Feature/node search backend and frontend

tsubik:
Global search everywhere (https://github.com/Vizzuality/trase/issues/197, https://github.com/Vizzuality/trase/issues/198, https://github.com/Vizzuality/trase/issues/199)

global search

Profile search is using API.

Various fixes:
- tool page search close when escape key pressed
- open tool page search if any valid key pressed
- some minor styling issues of tool page search 

Choosing backend branch as the base one to make easier to review frontend changes.


db/structure.sql
```diff
@@ -422,5415 +422,166 @@ ALTER SEQUENCE content.users_id_seq OWNED BY content.users.id;
 
 
 --
--- Name: ar_internal_metadata; Type: TABLE; Schema: public; Owner: -
```
tsubik: hmm, what happened here @tiagojsag ?

tsubik:
Adding "WIP" label as I need to clarify a few things on Monday, like the one below for example.

image

tiagojsag:
I removed the duplicates (@agnessa please review my solution) and added context identification to each result. One thing we do need is to group nodes like we do in the sankey search - using the example above, cargill exporter for brazil soy, and cargill importer for brazil soy should be a single result. As I mentioned, we already did this in the sankey, so reach out to @sorodrigo and @agnessa for details

tsubik:
The current state of things @juancarlosalonso 
image 2

PR #449, opened 2018-04-13
Make tool page pills always visible

tiagojsag:
Needs design review with JC

### DIFF
--- a/frontend/scripts/components/tool/nodesTitles.component.js
+++ b/frontend/scripts/components/tool/nodesTitles.component.js
@@ -33,13 +33,14 @@ export default class {
     if (nodesData === undefined || !nodesData.length) {
       return;
     }
-    // when map is full screen, show data as a tooltip instead of a nodeTitle
+    // if we have coordinates, request came from hover on map, so we have a tooltip and don't need to show pill
+    // else show pill for sankey node
     if (coordinates !== undefined) {
       this._showTooltip(nodesData, coordinates, currentQuant);
+    } else {
+      this.el.classList.remove('is-hidden');
+      this._update(!isHighlight, nodesData, recolorGroups, currentQuant, selectedYears);
     }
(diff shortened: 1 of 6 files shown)


sorodrigo:
After reviewing with juancarlos we came to the following conclusions:
A. Sankey columns and map-legend should be aligned at the bottom.
B. The pills container is shorter that it should be, scroll should happen when no more space is left. In the picture you can see 3 pills and 4 selected nodes.
C. Because the map has a tooltip there's no need to display the hovered polygon pill.

screen shot 2018-04-18 at 11 47 07
screen shot 2018-04-18 at 11 50 36

tsubik:
If we don't need the hovered pill to be displayed in the bar then I would also hide it when don't have any selected nodes. That would save up a little space for the map and the sankey.

tiagojsag:
A: cant find where that height is set, damn CSS
B: Extended the container, but don't know how to implement scroll, damn CSS
C: Done (because no damn CSS changes needed)

Tomasz, I thought of that too, but then we'd need a transition for something as simple as hovering a sankey node. I agree it does feel weird to see that blank space from the start, but since hovering over 40% of the screen will add data to it, I think it's best to leave it as it is

tsubik:
image

The scrollbar is visible now, but not in full, as everything there is done in absolute positioning/fixed height manner. I could make it fully visible but then pills would go up and there will be no spacing between them and sankey. 

1. I'm not sure if now displaying hovered sankey node details in the bottom-left of the screen is a good UX (it's soo far away). Maybe another tooltip there as for flows?
2. I think it would be great to highlight area on the map while hovering over pills.

tiagojsag:
Hey Tomasz,

I like both ideas, but they would be suggestions for next phase. c'ing @juancarlosalonso so he can take them into account.

thanks

Issue #420, opened 2018-04-05
Map fit bounds acts weird when municipality has no flows


sorodrigo:
When selecting and deselecting municipalities if you have at least one exporter selected then map fitbounds can do strange stuff.

How to reproduce:
1. Select Amaggi, Formosa do rio preto, Sao desiderio, Nova mutum. [Starting scenario](https://trase.earth/flows?state=eyJzZWxlY3RlZENvbnRleHRJZCI6MSwic2VsZWN0ZWRZZWFycyI6WzIwMTUsMjAxNV0sImRldGFpbGVkVmlldyI6ZmFsc2UsInNlbGVjdGVkTm9kZXNJZHMiOlsxODk3LDI1NTAsMTQyLDEwOTVdLCJleHBhbmRlZE5vZGVzSWRzIjpbXSwiYXJlTm9kZXNFeHBhbmRlZCI6ZmFsc2UsInNlbGVjdGVkQ29sdW1uc0lkcyI6WzMsNiw3LDhdLCJzZWxlY3RlZE1hcERpbWVuc2lvbnMiOlsicXVhbnQzNCIsbnVsbF0sImlzTWFwVmlzaWJsZSI6ZmFsc2UsIm1hcFZpZXciOnsibGF0aXR1ZGUiOi0xMi4xNzM3MzczMjU2OTUyODcsImxvbmdpdHVkZSI6LTUwLjY2MTIwOTM0MzEzNzMxLCJ6b29tIjo1fSwiZXhwYW5kZWRNYXBTaWRlYmFyR3JvdXBzSWRzIjpbXSwic2VsZWN0ZWRNYXBCYXNlbWFwIjoiZGVmYXVsdCIsInNlbGVjdGVkUmVzaXplQnlOYW1lIjoiVm9sdW1lIiwic2VsZWN0ZWRSZWNvbG9yQnlOYW1lIjoibm9uZSIsInNlbGVjdGVkQmlvbWVGaWx0ZXJOYW1lIjoibm9uZSJ9)
2. Deselect Formosa do rio preto.

tiagojsag:
can reproduce with instructions above :S

PR #455, opened 2018-04-17
Fixing fitting map to selected geo nodes and linked nodes

tsubik:
Fixing https://github.com/Vizzuality/trase/issues/420. If there are any selected linked nodes then map should not try to fit bounds to all selected nodes.


tsubik:
Should be better now @sorodrigo.

tsubik:
Just checked Rodrigo's case on the develop branch and this PR is still valid @tiagojsag. We agreed that if there are flows the map should fit the bounds to only those regions with flows.

PR #456, opened 2018-04-18
Feat: translate links

sorodrigo:
This PR includes behaviour to allow us to translate links urls. How does it work?
It will add a special transifex property to the `<a>` tags that need translating their href.

This is applied for document links only, so it's limited to the links in the home slider that point to documents and to the absolute links in the static pages.

We we're able to achieve this in the markdown pages thanks to glorious `remark-react` 🙌 . We provide a custom component for the `<a>` tags, and will use it to render all markdown links. This also allowed us to using `redux-first-router-link` `<Link />` for internal links.

Tested on transifex platform and the links are replaced by placeholders that need translating, so basically it _just works™_.

### DIFF
--- a/frontend/scripts/react-components/home/slider-section.component.jsx
+++ b/frontend/scripts/react-components/home/slider-section.component.jsx
@@ -76,7 +76,7 @@ class SliderSection extends React.PureComponent {
     const { className, name, slides } = this.props;
     const { visiblePages, currentSlide } = this.state;
     const smallScreen = visiblePages === 1;
-    const numColums = slides.length <= visiblePages && visiblePages < 3 ? 6 : 4;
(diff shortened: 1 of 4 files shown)


frontend/scripts/react-components/home/slider-section.component.jsx
```diff
@@ -76,7 +76,7 @@ class SliderSection extends React.PureComponent {
     const { className, name, slides } = this.props;
     const { visiblePages, currentSlide } = this.state;
     const smallScreen = visiblePages === 1;
-    const numColums = slides.length <= visiblePages && visiblePages < 3 ? 6 : 4;
+    const numColumns = slides.length <= visiblePages && visiblePages < 3 ? 6 : 4;
```
tiagojsag: <3

Issue #461, opened 2018-04-19
Change profile page summary texts for when actors/places have 0 flow


tiagojsag:
See https://trase.earth/profile-actor?nodeId=16508&lang=en

https://basecamp.com/1756858/projects/12498794/todos/334607473#comment_615194758

PR #464, opened 2018-04-20
Improve actor profile page when data is missing

tiagojsag:
https://github.com/Vizzuality/trase/issues/461


app/services/api/v3/actors/basic_attributes.rb
```diff
@@ -108,14 +112,23 @@ def importer_summary
...
+
           text += summary_of_sources('importer')
           text += summary_of_destinations('importer')
           text
         end
 
         def summary_of_total_trade_volume(profile_type)
-          initialize_trade_volume_for_summary
+          if @trade_total_current_year_raw.zero?
+            return "<span>#{@node.name.humanize}</span> \
```
agnessa: I only now realised we're returning html from the backend, this is horrible. I understand this is to support translations; in which case I propose to move the logic of constructing the summary text to frontend from values provided by backend. I'll create an issue for this.

PR #465, opened 2018-04-23
Update Rails to 5.1.6

tiagojsag:
Should fix the current dependency security issue


agnessa:
Build seems to be failing, I'm wondering whether it would be better to try upgrading just nokogiri here (which is the vulnerability) and see if it passes - if it does we need to dig deeper into the Rails changelog.

PR #467, opened 2018-04-24
Fix/expanded pill visibility

tsubik:
I've noticed that after my changes in the previous "re-expand branch" expand pill was not always visible when selecting node on the map or in the search. 
Another thing that this PR fixes is when redirecting to the tool using global search, selected nodes should be also expanded.


tsubik:
https://github.com/Vizzuality/trase/issues/469 fixed. Other improvements for global search is better to do on a separate branch.

PR #476, opened 2018-04-27
Feature/Google Analytics track not supported page

tsubik:
Couple things here:

- re-enable tracking of data portal events
- track not-supported page views (https://github.com/Vizzuality/trase/issues/472)

I wonder, should we enable pageview tracking for the rest of the pages?


tsubik:
I'm extending this to correctly track the visits for all of the pages.

tsubik:
Done. I've updated PR description with the new all pages tracking feature.

Issue #468, opened 2018-04-24
Improvements to search


tiagojsag:
Tiago
- [x] Hide map buttons on non subnational nodes
- [x] Load contexts on profile pages

Tomasz
- [x] Have a default action when clicking on a result -> take to sankey
- [x] Results lose highlight between result reloads
- [x] We have to make sure we only show the latest results/for current search term
- [x] Keep search content on blur (for example when user look at the other browser tab, and go back to the page)

PR #479, opened 2018-05-02
Global context state

tiagojsag:
This PR extract a few things into global app state:
- contexts list
- selected context
- tooltips

These 3 elements are now loaded once, on app bootstrap, and shared across the whole interaction with the frontend.

This required some changes a bit all over the place, more noticeably in the sankey state load/save.

Other changes I did in the process was to finally add page titles on each page - yay!

This fixes missing context info on profile pages: https://github.com/Vizzuality/trase/issues/468


frontend/scripts/actions/app.actions.js
```diff
@@ -17,6 +22,52 @@ export const TOGGLE_MAP_LAYERS_MENU = 'TOGGLE_MAP_LAYERS_MENU';
...
+    dispatch({
+      type: SET_CONTEXT_IS_USER_SELECTED,
+      payload: true
+    });
+
+    if (getState().location.type === 'tool') {
+      dispatch(loadToolDataForCurrentContext());
+    }
+
+    document.title = getPageTitle(getState());
```
tsubik: I wonder if that could be easily done somewhere in the router.

frontend/scripts/actions/tool.actions.js
```diff
@@ -82,6 +76,59 @@ const _reloadLinks = (param, value, type, reloadLinks = true) => dispatch => {
   }
 };
 
+const _setRecolorByAction = (recolorBy, state) => {
```
tsubik: Do we have some convention to use prefix `_` for private functions? I think as long is not exported it's clearly private.

frontend/scripts/actions/tool.actions.js
```diff
@@ -136,23 +183,15 @@ export function resetSankey() {
...
     });
 
+    const state = getState();
+
     if (defaultRecolorBy) {
-      dispatch({
-        type: SELECT_RECOLOR_BY,
-        name: defaultRecolorBy[0].name
-      });
+      dispatch(_setRecolorByAction({ value: defaultRecolorBy[0].name }, state));
```
tsubik: :+1:

frontend/scripts/pages/tool.page.jsx
```diff
@@ -15,14 +15,13 @@ import MapContextContainer from 'containers/tool/map-context.container';
 import MapLegendContainer from 'containers/tool/map-legend.container';
 import MapBasemapsContainer from 'containers/tool/map-basemaps.container';
 import MapContainer from 'containers/tool/map.container';
-import NavReactContainer from 'react-components/nav/filters-nav/filters-nav.container';
+import FiltersNav from 'react-components/nav/filters-nav/filters-nav.container';
```
tsubik: :100:

frontend/scripts/react-components/data-portal/data-portal-page.component.jsx
```diff
@@ -121,17 +134,16 @@ class DataContent extends Component {
         const selectedContext = this.props.contexts.find(
           elem => elem.isDisabled !== true && elem.id === value
         );
-        this.setState({ selectedCommodity: value, selectedContextId: selectedContext.id });
+        this.setState({ selectedCommodity: value });
+        this.props.selectContextById(selectedContext.id);
```
tsubik: Can `selectedContext` be null here? Also what is the difference between `props.selectContextById` and `props.onContextSelected` looks like duplication for me.

frontend/scripts/react-components/data-portal/data-portal.reducer.js
```diff
@@ -1,22 +1,13 @@
...
-  LOAD_CONTEXTS,
-  LOAD_EXPORTERS,
-  LOAD_INDICATORS
-} from 'actions/data.actions';
-import { createReducer } from 'store';
+import { LOAD_CONSUMPTION_COUNTRIES, LOAD_EXPORTERS, LOAD_INDICATORS } from 'actions/data.actions';
+import { createReducer } from 'scripts/store';
 
 const initialState = {
-  contexts: [],
```
tsubik: Great to remove that duplication :+1:

frontend/scripts/react-components/explore/explore.thunks.js
```diff
@@ -26,10 +36,13 @@ export const redirectToExplore = (dispatch, getState, { action }) => {
   if (toolPages.includes(action.type)) {
     if (!previouslyVisitedExplorePage.get()) {
       previouslyVisitedExplorePage.set(Date.now());
-      return dispatch(redirect({ type: 'explore' }));
+      dispatch({
```
tsubik: I saw an action creator function for that somewhere.

frontend/scripts/react-components/shared/app.thunks.js
```diff
@@ -0,0 +1,67 @@
...
+          type: SET_TOOLTIPS,
+          payload: data
+        });
+      })
+      .then(() => resolve())
+  );
+}
+
+function loadContextsPromise(dispatch, getState) {
+  // Contexts should only load once
```
tsubik: I see that on initial load the context is being loaded twice, probably because of the language selecting redirect. Maybe we need some isLoading state?

frontend/scripts/react-components/tool/tool.reducer.js
```diff
@@ -35,30 +31,34 @@ import {
...
-import getNodesDict from './helpers/getNodesDict';
-import getRecolorGroups from './helpers/getRecolorGroups';
-import getVisibleNodes from './helpers/getVisibleNodes';
-import mergeLinks from './helpers/mergeLinks';
-import setNodesMeta from './helpers/setNodesMeta';
-import sortVisibleNodes from './helpers/sortVisibleNodes';
-import splitLinksByColumn from './helpers/splitLinksByColumn';
-import splitVisibleNodesByColumn from './helpers/splitVisibleNodesByColumn';
+import { createReducer } from 'scripts/store';
+import filterLinks from '../../reducers/helpers/filterLinks';
```
tsubik: better to use absolute paths like `scripts/reducers/helpers/filterLinks`.

tiagojsag:
No, that's intentional. selected context is now shared accross the site

frontend/scripts/actions/app.actions.js
```diff
@@ -17,6 +22,52 @@ export const TOGGLE_MAP_LAYERS_MENU = 'TOGGLE_MAP_LAYERS_MENU';
...
+    const selectedContext = getContextById(getState(), contextId);
+
+    dispatch({
+      type: SET_CONTEXT,
+      payload: selectedContext
+    });
+
+    dispatch(setContextIsUserSelected(true));
+
+    dispatch({
```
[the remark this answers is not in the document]
tiagojsag: will fix

frontend/scripts/actions/app.actions.js
tiagojsag: Don't think so. The router only calls this when the page changes, which happens prior to this action.

frontend/scripts/actions/tool.actions.js
tiagojsag: we do, but we don't enforce it that much

sorodrigo:
> No, that's intentional. selected context is now shared accross the site

TBH I see this as an "anti-feature".

tiagojsag:
It felt a bit weird to me when I thought about it, but I also believe it helps have a more consistent experience while navitaging throught diferent pages. Overall, my main driver for this is that most users will be interested in a single context, and while we do carry that selection on certain interactions, this PR aims at also increasing the reach of that. 

I also discussed this with JC, and his comment was "can we also save it in a cookie, so we can load it even if the user leaves?". I haven't gone down that path yet, but I do feel like it is not a bad feature - even if it may feel confusing at first. perhaps it is so because we are used to the "old" way?

tiagojsag:
@tsubik both issues you reported should now be solved

tsubik:
I found another one. Unfortunately, it's a tricky one as sometimes it works and sometimes it does not. I'm not able to reproduce that all the times, but here are my steps.

1. Go to explore page, select for example Brazil - Beef. (there is another issue with date selector, default date is 2016, but that year is not available in selector)
2. Explore Supply Chain.
3. Change to Brazil - Soy.
4. Sometimes the choropleth layer is not showing up although Soy Production indicator is selected.

PR #490, opened 2018-05-21
Feat/indonesia ports new way of rendering

tsubik:
The new way to render single point layers such as Indonesia ports.

The size of the shadow marker is dependent on node's height on Sankey which is, of course, dependent on selected indicator's volume. 

screenshot from 2018-05-21 11-32-08
screenshot from 2018-05-21 11-32-35


tiagojsag:
- [ ] when I select a exporter/importer, the linked nodes on the map don't get colored yellow, and the map does not focus the sourcing ports

tsubik:
I fixed the issue and compared the behavior with the production. Also, based on designs, you were right, the point should be smaller and I made it 5px width.

PR #500, opened 2018-05-24
Transifex form fix

tsubik:
This PR fixes the download form which didn't play nice with Transfix live translations feature.

### DIFF
--- a/frontend/scripts/react-components/data-portal/data-portal-form.component.jsx
+++ b/frontend/scripts/react-components/data-portal/data-portal-form.component.jsx
@@ -20,49 +20,48 @@ class DataPortalForm extends Component {
     ];
 
     this.state = {
-      name: '',
-      country: '',
-      organisation: '',
-      organisationType: '',
-      dataUse: '',
-      comments: '',
-      email: '',
-      tos_check: '',
-
       showTOSWarning: false
     };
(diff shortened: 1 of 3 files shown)


frontend/scripts/react-components/data-portal/data-portal-form.component.jsx
```diff
@@ -114,8 +114,8 @@ class DataPortalForm extends Component {
                 We&apos;d love to hear about how you use the data and how we could improve it.
                 Please fill in details below and click on &apos;submit&apos;:
               </p>
-              <label htmlFor="name">
-                Name:
+              <label htmlFor="name" tx-content="exclude">
```
tiagojsag: could we move the input elements outside of the label element? this way, I think we are flagging the input as exclude, which I assume will result in the placeholder being ignored by transifex

tsubik: Alright, I managed to fix that issue in another way. The problem is that Transifex is re-adding input nodes on itself which is not friendly at all with React VDOM, but I think I have a working version, just need a few more minutes to test it and I will rebase that branch.

tsubik:
Done in another way @tiagojsag.

PR #502, opened 2018-05-25
Add links and vertix dot to sourcing adn destination charts on actor profile

tiagojsag:


frontend/scripts/react-components/profiles/line.component.jsx
```diff
@@ -173,35 +176,76 @@ class Line extends Component {
...
+                if (newNumberY + LINE_LABEL_HEIGHT > lastNumberY) {
+                  newNumberY = lastNumberY - LINE_LABEL_HEIGHT - 1;
+                }
+                lastNumberY = newNumberY;
+                return `translate(${width + 6},${newNumberY})`;
+              })
+
+              .text(`${numLines - i}.`);
+
+            texts
```
tsubik: not sure why not make the link along with the number. But if we want to have link only on the text then we can also group those texts together to not duplicate `transform` function.
```
const texts = pathContainers
  .selectAll('text')
  .data(d => [d])
  .enter()
  .append('g')
  .attr('transform', d => {
    const last = d.length - 1;
    const { value } = d[last];
    let newNumberY = y(value) + 4;
    if (newNumberY + LINE_LABEL_HEIGHT > lastY) {
      newNumberY = lastY - LINE_LABEL_HEIGHT - 1;
    }
    lastY = newNumberY;
    return `translate(${width + 6},${newNumberY})`;
  });

texts.append('text').text(`${numLines - i}.`);

texts
  .append('text')
  .attr('transform', 'translate(16,0)')
  .attr('class', d => {
    if (typeof onLinkClick !== 'undefined' && d[0].nodeId && d[0].profileType) {
      return 'link';
    }
    return '';
  })
  .on('click', d => {
    if (typeof onLinkClick !== 'undefined' && d[0].nodeId && d[0].profileType) {
      onLinkClick(targetLink, {
        nodeId: d[0].nodeId,
        year
      });
    }
  })
  .text(d => `${capitalize(i18n(d[0].name))}`);
```

tiagojsag: the line under the number looked meh. Will try your approach, thanks

app/services/api/v3/actors/top_nodes_summary.rb
```diff
@@ -84,9 +84,20 @@ def nodes_by_year_summary_for_indicator(
...
+            select('profiles.name').
+            first
+
+          profile_type_name = profile_type.name unless profile_type.nil?
+
           lines = @top_nodes.map do |node|
             {
               name: node['name'],
+              node_id: node['node_id'],
+              profile_type: profile_type_name,
```
agnessa: I believe it's overkill to include it with every "line", as the profile type is going to be the same for all lines; maybe consider moving it up to the same level as "unit" in the response.

PR #534, opened 2018-08-08
Test: profile page actor

sorodrigo:
This PR adds initial tests for the actor profile page. For testing, we're using puppeteer and jest which _just works_. We're mocking the API requests using puppeteer's built in request interceptor feature and Jest for assertions.


frontend/scripts/tests/puppeteer/profile-actor.test.js
```diff
@@ -0,0 +1,274 @@
...
+      await page.goto(`${baseUrl}/profiles`);
+      await page.waitForSelector('[data-test=profile-search]');
+      await page.click('[data-test=search-input-desktop]');
+      await page.type('[data-test=search-input-desktop]', nodeName);
+      await page.waitForSelector(`[data-test=search-result-${nodeType}-${nodeName}]`);
+      await page.click(`[data-test=search-result-${nodeType}-${nodeName}]`);
+
+      expect(page.url().startsWith(`${baseUrl}/profile-${profileType}`)).toBe(true);
+
+      expect.assertions(1);
```
aabdaab: do we need this? if anything in this example fails, the test itself will fail

frontend/scripts/tests/puppeteer/profile-actor.test.js
```diff
@@ -0,0 +1,274 @@
...
+    },
+    30000
```
aabdaab: what's the reason behind it? we shouldn't rely on wait times

frontend/scripts/tests/puppeteer/profile-actor.test.js
```diff
@@ -0,0 +1,274 @@
...
+            },
+            body: JSON.stringify(mocks[url])
+          }),
+        300
+      );
+    } else {
+      interceptedRequest.continue();
+    }
+  });
+});
```
aabdaab: I would move this whole method to some kind of helper and name it accordingly. You could have `helpers` folder for dealing with the configuration things for tests

aabdaab: it would be much more readable if it'd be something like:
```
beforeAll(... 
  openBrowser();
  openNewPage();
  setUpForTesting();
  ...
```
wdyt?

frontend/scripts/tests/puppeteer/profile-actor.test.js
```diff
@@ -0,0 +1,274 @@
...
+
+afterAll(() => {
+  browser.close();
```
aabdaab: this could also be moved to a helper like `helpers/destroy-app.js` or sth similar

frontend/scripts/tests/puppeteer/profile-actor.test.js
```diff
@@ -0,0 +1,274 @@
...
+
+describe('Profile actor', () => {
+  test(
+    'All 5 widget sections attempt to load',
+    async () => {
+      expect.assertions(1);
+
+      // await page.goto(`${baseUrl}/profile-actor?lang=en&nodeId=441&contextId=1&year=2015`);
+      await page.waitForSelector('[data-test=loading-section]');
+      const loadingSections = await page.$$('[data-test=loading-section]');
```
aabdaab: all the const you're creating here I would move to another file. Think of that kind of architecture:
`tests/pages/profile.js` where you put all the selectors
`tests/e2e/profile-actor.js` where you import the selectors from pages and have actual tests here, you could also split the tests for specific behaviours

frontend/scripts/tests/puppeteer/profile-actor.test.js
sorodrigo: Yeah it's weird thats after the awaits, it shouldn't be there. I guess that because it's only one, we can remove it

frontend/scripts/tests/puppeteer/profile-actor.test.js
sorodrigo: So 30 seconds it's the puppeteer timeout limit, and the tests take some time to be evaluated, more that the `jest.timeout` default. So I juust bumped it to the puppeteer timeout, so that if the test fails is because the browser timed-out and not jest.

frontend/scripts/tests/puppeteer/profile-actor.test.js
sorodrigo: I agree, started refactoring this on the `profile-place` tests.

frontend/scripts/tests/puppeteer/profile-actor.test.js
sorodrigo: I'm not sure about this one, I think it's clear enough. Maybe we could refactor it into a `destroy();` method but we'd still need to call it in each test's `afterAll`. It's a trade-off.

frontend/scripts/tests/puppeteer/profile-actor.test.js
sorodrigo: I'm wary of pulling thing up too soon, but I agree it doesn't make sense repeating the selector. Perhaps just hoisting this to the top level would do the trick. 👍

sorodrigo:
@aabdaab Thanks! I think in general I agree with your suggestions, I will make some changes to clean it up, and some other might come later in the `profile-place` tests. Let me know what you think about the timeout limit thingy.

frontend/scripts/tests/puppeteer/profile-actor.test.js
aabdaab: in that case maybe you can create `cleanTest` function that wraps up test function and adds the timeout limit and use that one instead of `test`

aabdaab: and `cleanTest` could land in the helpers folder! :D

PR #542, opened 2018-08-22
Feature/configurable profile pages

agnessa:
This is in progress.

Reduces hardcoded references to indicators in the place / actor endpoints.

`bundle exec rake db:migrate`
`bundle exec rake charts:reload`

The configuration of profile pages is stored in the database, using the following tables:

<pre>                                                                                         
                                                                            ┌────────────────┐
                                                                            │                │
                                                                      ┌────┼│   chart_inds   │
                                                                      │     │                │
                                                                      │     └────────────────┘
                                                ┌────────────────┐    │     ┌────────────────┐
                                               ╱│                │    │     │                │
                                        ┌───────│chart_attributes│┼───┼────┼│  chart_quals   │
                     ┌─────────────┐    │      ╲│                │    │     │                │
┌─────────────┐     ╱│             │    │       └────────────────┘    │     └────────────────┘
│  profiles   │┼───○─│   charts    │┼───│                             │     ┌────────────────┐
└─────────────┘     ╲│             │    │       ┌────────────────┐    │     │                │
                     └─────────────┘    │      ╱│                │    └────┼│  chart_quants  │
                                        └───────│chart_node_types│          │                │
                                               ╲│                │          └────────────────┘
                                                └────────────────┘</pre>

A chart represents a grouping of attributes used in a particular chart on the profile page.  It has an `identifier` property, which is referenced in the API code to retrieve the attributes.

`inds`, `quals` and `quants` can be linked to a chart via `chart_inds/quals/quants`, each of which belongs to a single chart_attribute. `chart_attributes` table contains additional visualisation properties, such as display_name, legend_name and position. Where more than one attribute is assigned to a chart, e.g. a set of attributes to be displayed in a table, they need to be assigned position in which they apper. Where an attribute is used in a specific function, it needs to have an identifier which is referenced in the code, e.g. the attribute for commodity production set to `SOY_TN`. There's a materialised view `chart_attributes_mv` which brings all that together, analogically to how we handle `map/download/resize_by/recolor_by_attributes`.

Node types can be linked to a chart through `chart_node_types`. Similarly to attributes, they can have position (e.g. in a chart where we show data by `BIOME`, `STATE`, `MUNICIPALITY` - these three node types should be assigned position according to where they appear). They can also have an identifier, which is referenced in the code to use the correct node type in the correct place, e.g. setting the node type for trade destination to `COUNTRY`.
### Tasks:
- [x] populate charts and chart attributes with current configuration
- [x] remove hardcoded references to indicators:
  - [x] PLACE:
    - [x] basic attributes (depends on SOY_TN, SOY_YIELD, SOY_AREA_PERC, AREA_KM2)
    - [x] sustainability indicators table, which is represented in the database as 4 separate charts (each corresponding to one tab): `:environmental_indicators`, `:socioeconomic_indicators`, `:agricultural_indicators`, `:territorial_governance` (attributes for the chart)
    - [x] deforestation trajectory `:trajectory_deforestation` (attributes for the chart)
  - [x] ACTOR:
    - [x] basic attributes
    - [x] sustainability indicators table `:sustainability`
    - [x] exporting companies `:companies_sourcing`
    - [x] top destinations (depends on SOY_TN)
    - [x] top sources (depends on SOY_TN)
  - [x] consider whether charts which only depend on `Volume` require this kind of configuration (place / top traders, place / top consumers, actors / sustainability table, actors / exporting companies, actors / top sources, actors / top destinations) _Volume stays hardcoded, it's a special type of attribute that would have to be configured identically for every context, so that would not be helpful_
- [x] remove hardcoded references to place node types (MUNICIPALITY, STATE, BIOME, COUNTRY)
  - [x] place / basic attributes
  - [x] place / top consumers
  - [x] actors / basic attributes
  - [x] actors / sustainability table
  - [x] actors / top sources
  - [x] actors / top destinations
  - [ ] sustainability indicators and deforestation trajectory uses the STATE qual to produce state ranking; check what happens when it is missing
- [x] remove hardcoded references to actor node types (EXPORTER, IMPORTER)
  - [x] place / basic attributes
  - [x] place / top consumer actors
  - [x] actor / basic attributes
- [x] admin pages to manage charts and chart attributes
- [x] consider whether summary should continue to be generated in the backend _Probably better frontend, but not for now_
- [ ] some labels are hardcoded in the frontend, for example chart titles


tiagojsag:
dont merge - delay until after we release

PR #544, opened 2018-08-23
Prevent data import deadlocks

tiagojsag:


agnessa:
In addition the build breaks because the views need to be populated 🤦‍♀️

PR #545, opened 2018-08-27
Feat: multiple contexts in profiles

sorodrigo:
This PR tweaks the context-selector to decouple it as much as possible from the store. Now I believe it's easier to reuse. I moved some props it received from the store to the parent components, this way this props can change depending where it's being used.

Finally, I added the context selector to the profile-root page, showing only contexts that have profile pages. Mobile styles included.

The next step is to connect it to the search and test it works fine.


sorodrigo:
Ready for reviewing 😄

tiagojsag:
image

This needs some designer love:
- letters feel too big
- The "search a company or production place" is faded, but you can't really tell until you actually select a context and see it "un-fade". We should improve this.

PR #561, opened 2018-10-08
Add more node types as destinations to dashboards_flow_paths mv; Fix …

tiagojsag:
Add more node types as destinations to dashboards_flow_paths mv; Fix rubocop execution through git hooks

### DIFF
--- a/db/structure.sql
+++ b/db/structure.sql
@@ -1513,7 +1513,7 @@ CREATE MATERIALIZED VIEW public.dashboards_flow_paths_mv AS
     cnt_props.column_group,
         CASE
             WHEN (cnt_props.column_group = 0) THEN 'SOURCE'::text
-            WHEN (node_types.name = 'COUNTRY'::text) THEN 'DESTINATION'::text
+            WHEN (cnt_props.column_group = 3) THEN 'DESTINATION'::text
             ELSE 'COMPANY'::text
         END AS category
    FROM (((((( SELECT flows.context_id,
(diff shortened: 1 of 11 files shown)


db/structure.sql
```diff
@@ -1119,6 +1126,202 @@ CREATE SEQUENCE public.country_properties_id_seq
 ALTER SEQUENCE public.country_properties_id_seq OWNED BY public.country_properties.id;
 
 
+--
+-- Name: dashboard_template_commodities; Type: TABLE; Schema: public; Owner: -
```
agnessa: I think we have stuff that doesn't belong here? One way to keep this file limited only to changes introduced in current branch is to 1. reset it to be a copy of base branch 2. `RAILS_ENV=test rake db:drop db:create db:structure:load` 3. `RAILS_ENV=test rake db:migrate`

tiagojsag: Yeah, I went down the lazy path, as always :P will do that and push my changes

tiagojsag:
There, all laziness removed :muscle:

db/views/dashboards_flow_paths_mv_v02.sql
```diff
@@ -0,0 +1,30 @@
...
+    a.node_id,
+    a."position"
+  FROM flows, LATERAL unnest(flows.path) WITH ORDINALITY a(node_id, "position")
+) flow_paths
+JOIN contexts ON flow_paths.context_id = contexts.id
+JOIN nodes ON flow_paths.node_id = nodes.id
+JOIN node_types ON nodes.node_type_id = node_types.id
+JOIN context_node_types cnt ON node_types.id = cnt.node_type_id AND flow_paths.context_id = cnt.context_id
+JOIN context_node_type_properties cnt_props ON cnt.id = cnt_props.context_node_type_id
+WHERE cnt_props.column_group = 0 OR node_types.name IN ('COUNTRY', 'IMPORTER', 'EXPORTER', 'TRADER');
```
agnessa: sorry I missed this - this condition needs to be changed as well to retrieve correct row: `WHERE cnt_props.column_group IN (0,3) OR node_types.name IN ('IMPORTER', 'EXPORTER', 'TRADER');`

tiagojsag:
@agnessa ready for re-re-review

PR #568, opened 2018-10-11
Feat/dashboards kickoff tickets

MarcoAntonioAntunes:
Added cookie feature into the create your own dashboard.

### TEST####

While creating your new dashboard:

- First visit should go to Welcome panel.
- Second visit or if pop-modal is closed it should jump automatically to the second panel.

### DIFF
--- a/frontend/scripts/react-components/dashboard-element/dashboard-element.container.jsx
+++ b/frontend/scripts/react-components/dashboard-element/dashboard-element.container.jsx
@@ -18,6 +18,8 @@ const mapDispatchToProps = dispatch =>
     dispatch
   );
 
+let hasVisitedBefore = localStorage.getItem('hasVisitedBefore');
+
 class DashboardElementContainer extends React.Component {
   static propTypes = {
     goToRoot: PropTypes.func.isRequired,
@@ -27,8 +29,15 @@ class DashboardElementContainer extends React.Component {
   state = {
     modalOpen: true,
     goBackOnCloseModal: true,
-    step: DashboardElement.steps.WELCOME
+    step: hasVisitedBefore ? DashboardElement.steps.PANEL : DashboardElement.steps.WELCOME
   };
+
+  componentDidMount() {
+    if (!hasVisitedBefore) {
+      hasVisitedBefore = true;
+      localStorage.setItem('hasVisitedBefore', true);
+    }
+  }
 
   closeModal = () => {
     this.setState({ modalOpen: false });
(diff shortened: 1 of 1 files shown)


frontend/scripts/react-components/dashboard-element/dashboard-element.container.jsx
```diff
@@ -27,9 +29,16 @@ class DashboardElementContainer extends React.Component {
   state = {
     modalOpen: true,
     goBackOnCloseModal: true,
-    step: DashboardElement.steps.WELCOME
+    step: hasVisitedBefore ? DashboardElement.steps.PANEL : DashboardElement.steps.WELCOME
   };
 
+  componentDidMount() {
+    if (!hasVisitedBefore) {
+      hasVisitedBefore = true;
```
sorodrigo: The reason you need to set the hasVisitedBefore to true is because this value is read on `dashboard-element.container.jsx` import and not on instantiation. This is because the value is initialised out of the class. If you move the hasVisitedBefore into the class as a class property (just like `state`) then there's no need to update the value to true because the value will be read everytime the component is instantiated.

frontend/scripts/react-components/dashboard-element/dashboard-element.container.jsx
```diff
@@ -18,6 +18,8 @@ const mapDispatchToProps = dispatch =>
     dispatch
   );
 
+let hasVisitedBefore = localStorage.getItem('hasVisitedBefore');
```
sorodrigo: Let's use a more representative name, something like: `TRASE__HAS_VISITED_DASHBOARDS_BEFORE`.

frontend/scripts/react-components/dashboard-element/dashboard-element.container.jsx
```diff
@@ -27,9 +29,16 @@ class DashboardElementContainer extends React.Component {
...
     modalOpen: true,
     goBackOnCloseModal: true,
-    step: DashboardElement.steps.WELCOME
+    step: hasVisitedBefore ? DashboardElement.steps.PANEL : DashboardElement.steps.WELCOME
   };
 
+  componentDidMount() {
+    if (!hasVisitedBefore) {
+      hasVisitedBefore = true;
+      localStorage.setItem('hasVisitedBefore', true);
```
sorodrigo: Finally I think a good practice is to save a date timestamp (`Date.Now()`) instead of a boolean. This allows us to reuse the same cookie in the future for users that have seen the original modal, maybe showing a new message.

tiagojsag:
`git rebase develop`

MarcoAntonioAntunes:
Applied the necessary changes.

PR #569, opened 2018-10-11
Load dashboard templates as frontend cards

tiagojsag:
Missing (for other PRs)
- create the actual links to the dashboards
- links to indicators

### DIFF
--- a/app/admin/api/v3/dashboard_template.rb
+++ b/app/admin/api/v3/dashboard_template.rb
@@ -112,4 +112,7 @@ ActiveAdmin.register Api::V3::DashboardTemplate, as: 'DashboardTemplate' do
       end
     end
   end
+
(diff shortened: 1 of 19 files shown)


tiagojsag:
ready for review

app/admin/api/v3/dashboard_template.rb
```diff
@@ -16,6 +18,7 @@ def clear_cache
     inputs do
       input :title, as: :string, required: true
       input :description, required: true
+      input :category
```
agnessa: is this free text, or possible to have an enumeration with possible values in a dropdown?

tiagojsag: AFAIK there's no definition on that, so I'm leaving it open (and optional) for now.

agnessa: 👌

PR #572, opened 2018-10-11
Feat/dashboard welcome

MarcoAntonioAntunes:
Built welcome dashboard based on the designs
Added the images and functionality on the second panel

###TEST###

- Go through both welcome and second panel - compare with the designs
- test functionality


frontend/styles/components/dashboard-element/dashboard-welcome.scss
```diff
@@ -1,27 +1,67 @@
...
 
   .dashboard-welcome-card-container {
     display: flex;
-    margin: 0 0 30px 0;
+    margin: 0 0 60px 15px;
   }
 
   .dashboard-welcome-card {
-    background-color: darken($light-gray, 7%);
+    font-family: Merriweather;
```
sorodrigo: Let's use the sass variables when setting the `font-family` and `font-weight` values. For other things such as `font-size` and `margin` and `padding` we aren't really using the sass variables anymore. I don't really expected you to know this so let's discuss it later.

frontend/styles/components/dashboard-element/dashboard-welcome.scss
```diff
@@ -1,27 +1,67 @@
...
+      font-style: normal;
+      font-stretch: normal;
+      line-height: 1.6;
+      letter-spacing: normal;
+      color: #34444c;
+      margin-bottom: 10px;
+    }
+
+    img {
+      object-fit: contain;
```
sorodrigo: `object-fit` is not supported by IE 11 😭 (https://caniuse.com/#search=object-fit) so we need to use a div with `background-image` instead. With a div, positioning the image should be easier also.

MarcoAntonioAntunes: I removed object-fit since it was being used unnecessarily.

MarcoAntonioAntunes:
Fixed and updates are now complete. Please review and merge.

PR #595, opened 2018-10-22
Make dashboards_flow_path_mv refresh run in the background

tiagojsag:


app/models/api/v3/readonly/dashboards/flow_path.rb
```diff
@@ -35,6 +35,10 @@ def self.refresh_dependents(options = {})
               mview_klass.refresh(options.merge(skip_dependencies: true))
             end
           end
+
+          def self.long_running?
```
agnessa: I'd add the same to FlowAttribute and NodeAttribute, because if it's refreshed "with dependents", and dependents are processed synchronously, then we might end up in an inconsistent state

tiagojsag: you mean long_running? set t true? they already have it.

PR #609, opened 2018-10-24
Feature/bar chart title units

weberjavi:
This PR adds a top label to the `dashboard-widget` with the type of indicator displayed and its unit.
[Pivotal task](https://www.pivotaltracker.com/story/show/161215968)
[Designs](https://projects.invisionapp.com/share/9GNS32PZYW3#/screens/317014652)

image
image

### DIFF
--- a/frontend/scripts/react-components/chart/axis-label/axis-label-component.jsx
+++ b/frontend/scripts/react-components/chart/axis-label/axis-label-component.jsx
@@ -1,0 +1,14 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const AxisLabel = props => {
+  const { text, suffix } = props;
+  return <p className="widget-yAxis-label">{`${text} [${suffix}]`}</p>;
+};
+
+AxisLabel.propTypes = {
+  text: PropTypes.string.isRequired,
+  suffix: PropTypes.string.isRequired
+};
+
+export default AxisLabel;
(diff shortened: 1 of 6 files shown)


frontend/scripts/react-components/chart/axis-label/axis-label-component.jsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const percentageSuffix = (text, suffix) => `${suffix} ${text}`;
+const regularSuffix = (text, suffix) => `${text} [${suffix}]`;
+
+const AxisLabel = props => {
```
sorodrigo: I think this should live along with the rest of the dashboard-widget components, the implementation is coupled to what the dashboard widget endpoint returns.

frontend/scripts/react-components/chart/axis-label/index.js
```diff
@@ -0,0 +1,3 @@
+import Component from './axis-label-component';
```
sorodrigo: I understand you made this while following the convention inside the `chart/` directory; makes total sense, however if we move this inside the `react-components/dashboard-element/dashboard-widget` directory we should follow the convention that it's used there. Also regarding file names.

I know it's not great, but I prefer having everything follow the same convention until we get to improve it. The `chart/` is an exception because I copy-pasta from another project, and eventually should leave as an independent module _(wishful thinking  someone_ 🙄  _? )._

weberjavi: Yes, you're right, I started making the label part of the chart (as `tick`) but after some back and forth I ended up extracting it. My mistake. Will adapt it to follow the convention of `dashboard-widget`'s components.

sorodrigo:
Thanks! I forgot to mention another thing, so I just added myself. 😛

PR #612, opened 2018-10-24
Return summary of selected filters as y value label

agnessa:
e.g. 'BR / SOY / BUNGE'


tiagojsag:
no tests? :(

agnessa:
I already forgot what this was for, would it still be useful @sorodrigo or will it break things if we add tests and merge?

sorodrigo:
It won't break stuff but designs changed so it won't be used :/

agnessa:
Out with it then!

PR #618, opened 2018-10-25
Fix: refactor dashboard panel reducer

sorodrigo:
This PR refactors the `dashboard-panel.reducer`. I removed the adhoc variable names in favor of calling every item prop as `activeItem` and every tab prop as `activeTab`. Also extracted countries into a new panel, this is actually an abstract panel because there's no UI for it, but makes way more sense data wise.

I refactored some of the actions and components to fit to this new structure and removed a saga that set the first tab as active. Now instead of using a saga, we do this a reducer level. This solution seems more solid.

Finally the big WHY behind this refactor? I'm paving the road for my next PR that will search nodes across all tabs of the same panel and if a node is selected and doesn't belong to that tab, the tab will be changed.

Have fun 🌮

### DIFF
--- a/frontend/scripts/react-components/dashboard-element/dashboard-element.actions.js
+++ b/frontend/scripts/react-components/dashboard-element/dashboard-element.actions.js
@@ -107,12 +107,12 @@ export const setDashboardActivePanel = activePanelId => ({
   payload: { activePanelId }
 });
 
-export const setDashboardPanelActiveItem = ({ activeItem, panel }) => ({
+export const setDashboardPanelActiveItem = (activeItem, panel) => ({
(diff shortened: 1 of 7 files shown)


sorodrigo:
Thanks I fixed that and several other hard to find hard to reproduce bugs. Ready for reviewing! :v:

weberjavi:
Still experiencing some inconsistencies when selecting`sourcing places` changing panel and going back to sources (just the selected country is displayed).
Also after changing panel there are some weird behaviours (unwanted deselections).
Please see the `gif`s attached.

kapture 2018-10-26 at 13 17 27

kapture 2018-10-26 at 13 07 38

PR #627, opened 2018-10-31
Added countryId to companies endpoint

agnessa:


tiagojsag:
feels like a test's missing here

agnessa:
> feels like a test's missing here

Tiago is chasing me about tests - my job is done here, my life is complete! 💃 🎉 ❤️

agnessa:
The endpoint is already covered by a schema spec, so I just updated the schema with the new property.

tiagojsag:
ok, so there's probably an issue in those tests, as they were not failing before you updated them.

agnessa:
Added countryId to companies search endpoint (also sources and destinations as all 3 use shared functionality and it was easier to do that than differentiate)

sorodrigo:
Ready to go!

PR #659, opened 2018-11-29
Feature/new transport mechanism for data uploader

agnessa:
https://www.pivotaltracker.com/story/show/162153451

Support for an alternative method of loading new data for the database update process. Old mechanisms left in place, most of the functionality is shared.

The new process has an extra step at the beginning - downloading a data dump from S3 and restoring it into the database. 
It's important that:
* this is a schema with data (tables), not materialised views, and has no references to other schemas (see [here](https://basecamp.com/1756858/projects/12498794/messages/81835847#comment_664470581))
* the name of the schema is not one of `public`, `content`, `main`; currently using `tool_tables` for that schema (which is configurable via env var - update to `.env` needed)
* the dump contains just this schema, e.g. `pg_dump -n tool --no-owner -U postgres dbname | gzip > data_20181127.gz`
* the dump is gzipped and its name ends in `.dump.gz` (because that's what we use for the full dump import feature and I reused it here)
* that the dump is saved in the same S3 bucket as the full database dumps, in the `MAIN` folder

A sample dump for testing is present in S3. You can test using the CMS (going to the existing database update page and selecting `MAIN/tool_tables_20181128.dump.gz` from the list) or using the new rake task: `SCHEMA_VERSION=MAIN/tool_tables_20181128.dump.gz bundle exec rake db:mirror:import`

Once the schema is restored, the import process proceeds exactly like before with replacing content of blue tables and restoring yellow tables, except now it runs locally rather than over postgres_fdw remote connection, which makes it much faster.

The slowest part is now refreshing the materialized views. I have an idea about fixing it by always refreshing the mviews in background; the downside is that this leads to inconsistent data while the mviews are refreshing.


tiagojsag:
dont merge - delay until after we release

PR #662, opened 2018-12-03
Fix: missing area line charts

sorodrigo:
This PR fixes a regression, where area line paths were not being renderer. It also updates the e2e tests to prevent this from happening again.


tiagojsag:
tests <3

PR #670, opened 2018-12-21
Pre-computed downloads

agnessa:
Modification of the download service so that in case of full downloads it first checks if a file is already generated and serves that; in case it's not, after generating it it stores it for future use.

After the download flows materialised view refreshes, it precomputes all full pivot csv downloads.

There are capistrano tasks:

`cap [env] downloads:clear`
`cap [env] downloads:refresh` <- this one clears immediately and refreshes in background after `sidekiq:restart`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,7 @@
 !/log/.keep
 !/tmp/.keep
 
+/public/downloads
 /public/system
 /public/urls.txt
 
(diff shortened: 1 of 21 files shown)


agnessa:
TODO:
consider other scenarios for clearing, e.g.
- capistrano task (not necessarily part of regular deploy)
- ensure clearing after db import (should already be in place, but check)
- button in the admin CMS +  warnings to remind humans to clear that
- check about varnish; does it even cache those downloads? if so we need clearing

agnessa:
* added a Capistrano task that runs after `sidekiq:restart` (because it refreshes asynchronously)
* added a Varnish BAN when clearing
* did not add an admin button to clear; workaround for admin is to update something in download attributes

Ready for review @tiagojsag

PR #674, opened 2019-01-07
Update charts config seeds to match the latest prod db

tiagojsag:


frontend/scripts/utils/formatValue.js
```diff
@@ -2,7 +2,7 @@ import { NUM_DECIMALS, NUM_DECIMALS_DEFAULT } from 'constants';
 
 // returns a value rounded to numDecimals
 export default (value, dimensionName) => {
-  if (value === undefined || value === null) {
+  if (value === undefined || value === null || value === 'NaN') {
```
sorodrigo: value being a string `'NaN'` is weird. I think it's a code smell. If anything it should be plain `NaN` and use 'Number.isNaN` to check it.

frontend/scripts/react-components/profiles/place-summary.component.jsx
```diff
@@ -52,9 +52,9 @@ class PlaceSummary extends React.PureComponent {
         onValueSelected: onYearChange
       }
     ];
-    const soyValue = soyArea !== null && soyArea !== 'NaN' ? formatValue(soyArea, 'area') : '-';
```
tiagojsag: @sorodrigo the NaN bit was copy-pasted from this

agnessa:
approved, just this one style violation

frontend/scripts/react-components/profiles/place-summary.component.jsx
sorodrigo: Yeah it’s out of the scope of this PR. Maybe just add a ticket to PT?

Issue #551, opened 2018-09-27
Validation report: warn when path length different than number of declared context node types


agnessa:
- Warn on post-import validation when path length different than number of declared context node types
- Warn on post-import validation when number of declared context node types < 4
- Crash import process when path length on any flow is lower than 4

PR #679, opened 2019-01-11
Validate paths length matches context node types and is minimum 4 nodes

agnessa:
https://www.pivotaltracker.com/story/show/162124214

https://github.com/Vizzuality/trase/issues/551

This adds a new validation to help with issues resulting from incorrect structure of flow paths. The `flows` table describes the path through which a particular commodity travelled. It is a denormalized table, where the `path` column is an array of node ids. The length of that array depends on the context and should match the number of records in `context_node_types`. For practical reason the path should never be shorter than 4 nodes (because otherwise we cannot render the sankey correctly).

This PR adds
* a `CHECK` constraint on the flows path to ensure it is never shorter than 4 nodes. Adding this as a constraint on the database will crash the import process early.
* a validation that compares the number of nodes declared in `context_node_types` with the length of path for this context.

### DIFF
--- a/app/services/api/v3/database_validation/chain_builders/context_chain_builder.rb
+++ b/app/services/api/v3/database_validation/chain_builders/context_chain_builder.rb
@@ -36,6 +36,7 @@ module Api
           checks :has_at_least_one,
                  association: :contextual_layers,
                  link: :index
+          checks :path_matches_context_node_types
           checks :active_record_check, on: :context_property, link: :edit
 
           CHAIN_BUILDERS = [
--- a/app/services/api/v3/database_validation/checks/path_matches_context_node_types.rb
+++ b/app/services/api/v3/database_validation/checks/path_matches_context_node_types.rb
@@ -8,9 +8,10 @@ module Api
           # @return (see AbstractCheck#passing?)
           def passing?
             expected_length = @object.context_node_types.count
-            @violating_flows_cnt = @object.flows.
-              where('ICOUNT(path) <> ?', expected_length).count
-            @violating_flows_cnt.zero?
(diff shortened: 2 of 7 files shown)


agnessa:
Yes, we could say there are two types of validations here, one expressed as a db constraint and one as a custom validator.

Database constraints are preferred most of the time, because they allow to crash the import process early. The import process is essentially restoring data from a dump file, so in that context regular AR validations don't run.

Database constraint approach is not always practical to apply, e.g. with the validation that checks the number of context node types vs path length, all records need to be in place at the time it runs and there's no control over the order in which records are created. In those cases the custom validators were developed, which don't tap into Active Record lifecycle, which often gives a performance benefit, e.g. instead of validating each flow, it runs a query on all flows per context to check for offenders.

What made me think is your earlier suggestion to use a custom AR callback `:data_validation` and this way keep the validations with the models without affecting how the CMS works; I think that's a great idea. Now it seems to me that would be much nicer if I had thought about it earlier, as domain logic would not be scattered all over the place. In cases like this, I would add the validations to the context rather than to the flow, so we could still use batch checks rather than record by record. I'm worried though about the time it will take me to refactor all validations this way, as one thing worse from badly structured code is badly structured and inconsistent code 😱 So I probably need to skip it for now.

tsubik:
Yeah, that would be time consuming and also more research will be needed to see if everything what those custom checks give is possible to implement using different way.

tsubik:
Ok, I tested the validation and it works :+1: 

I haven't tested importing any wrong database.

One suggestion though. It does not have to be implemented if not necessary. Instead of having the number of violating flows, maybe it's better to have their ids?

PR #686, opened 2019-01-17
Check for and destroy widows after import

agnessa:


tsubik:
Checking... 

Just a quick thing after reading the description. I never encountered word "widow" in this context. I would ersonally use "orphan".

agnessa:
> Checking...
> 
> Just a quick thing after reading the description. I've never encountered word "widow" in this context. I would personally use "orphan".

That's a good point. Most usually "orphan" refers to a record whose "parent" is gone without cleanup of the foreign key, which is not the case here - here all "children" are gone. I don't remember anymore where I got "widow" from, I might have read it somewhere a long time ago, or possibly applied typesetting terminology to relations 😅, but sure enough I cannot find a legitimate reference anymore, so will change, thanks!

app/services/api/v3/import/importer.rb
```diff
@@ -80,6 +81,51 @@ def import
...
+            cnt_after = table_class.count
+            next unless cnt_after != cnt_before
+
+            update_yellow_table_stats_after(table_class, cnt_after)
+          end
+        end
+
+        def update_yellow_table_stats_after(table_class, cnt_after)
+          table_name = table_class.table_name
+          blue_table_name = ALL_TABLES.find do |table|
```
tsubik: I think that would not return table_name but the whole hash, the table name still will be in `table[:table_class].table_name`

tsubik: That could be then shorten to 
```
blue_table_name = ALL_TABLES.find { |t| { t[:yellow_tables]&.include?(table_class) }&.dig(:table_class)&.table_name
```
not exactly in one line if too long

tsubik:
Ok, so I tested and the not referenced attribute has been removed after the import/update. As I thought, stats were not correctly updated.

app/models/concerns/api/v3/associated_attributes.rb
```diff
@@ -18,6 +18,12 @@ def manage_associated_attributes(ary)
         def associated_attributes
           @associated_attributes
         end
+
+        def destroy_zombies
```
tsubik: Previously on AMC's The Walking Dead... :heart:

PR #696, opened 2019-01-31
Fix/click through arrows

agnessa:
Adds a `profile` flag to inform if profiles exist for given node type. Also fixes funny capitalisation.

https://www.pivotaltracker.com/story/show/163590977
https://www.pivotaltracker.com/story/show/163591074

screenshot 2019-01-31 at 13 00 43


sorodrigo:
@agnessa 
Ok I fixed the profile-actor table links to the profile-place page. I found a similar case in the `profile-place` page. We have links in the mini-sankey that go to companies, but in the countries we don't. Currently we achieve this by either passing a page to the mini-sankey or not passing it. Do you think we should also handle this on the backend given that country profiles are coming? Or should we address that when we start the country profiles?

agnessa:
I wouldn't fix it now.

sorodrigo:
Ready for reviewing!

sorodrigo:
I added a regression test, will merge!

PR #698, opened 2019-02-05
fixes the recolorby order and colors

sorodrigo:
This PR fixes the order of the biome recolorby filter colors. I changed it so it matches production, please let me know if this works 🙏 .


agnessa:
Still looks different to me (both biome and zero deforestation options), should I have deleted some files to observe the change?

sorodrigo:
This won't fix things, it actually doesn't change a line of code, but it formats the code so hey that's something: merging...

PR #699, opened 2019-02-06
Fix/optimise actor top sources

agnessa:
https://www.pivotaltracker.com/story/show/162153340

replaces element lookup in an array (O(n)) with a hash (O(1)), which seems to speed up things a lot (~ 10-fold decrease in response time and increase in transaction rate for actors top sources endpoint)

### DIFF
--- a/app/services/api/v3/actors/top_nodes_summary.rb
+++ b/app/services/api/v3/actors/top_nodes_summary.rb
@@ -72,12 +72,19 @@ module Api
             include_domestic_consumption: false,
             limit: nil
           )
-          @top_node_values_by_year = top_nodes_list.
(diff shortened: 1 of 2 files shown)


agnessa:
@aabdaab if you have a moment 🙏 The actor profile has a number of charts on it and data for each chart is fetched from a separate endpoint. This PR is about optimising the endpoint that populates the "top sourcing regions".

Example profile page: https://sandbox.trase.earth/profile-actor?nodeId=32228&contextId=1&year=2017
top sources endpoint: https://sandbox.trase.earth/api/v3/contexts/1/actors/25038/top_sources?year=2017

PR #700, opened 2019-02-06
Update tabs component with new styles, use Text and update the styles…

Bluesmile82:
This PR updates the Tabs component with the new styles

image

The one in the designs has 14px font size but I used 13px as is not a standard size that we have. We can changed if its accepted by design.

- Also Text component is used in tabs and some other places
- New color 'egg' is added to text as we needed it for the Tabs component

After merging the Docz PR we should include it there


sorodrigo:
Thanks for adding `<Text />` where necessary!

PR #703, opened 2019-02-12
Fix/print mode and tests

Bluesmile82:
This PR fixes the print mode and creates some tests for it:

- Fix problem with visible tabs in print mode
- Add jest-image-snapshot and tests
- Update readme
- Add PDF file extension to the downloaded file

Note: There is still a problem with the iframe coming from GFW. It should be styled in GFW


frontend/styles/_print.scss
```diff
@@ -22,4 +22,10 @@
   .page-break-after-avoid {
     page-break-after: avoid;
   }
+
+  .c-tabs {
```
sorodrigo: I know I told you to put the styles here, but actually it made more sense inside the profile-actor and profile-place layouts, we are adding page specific print styles there. I made the change myself.

README.md
```diff
@@ -138,6 +138,30 @@ If you are using the included development server, you can set those variables in
 
 [![Build Status](https://travis-ci.org/Vizzuality/trase.svg?branch=master)](https://travis-ci.org/Vizzuality/trase)
 
+We use jest and puppeteer for the end to end tests. There are some mocks for the requests in the puppeteer/mocks folder. Every slash (/) and params (? &) are replaced by a hyphen in the file name. i.e:
```
sorodrigo: Thanks for adding docs!!

PR #705, opened 2019-02-15
Sorting out a few things in the data uploader

agnessa:
A few things:
- delete the temporary schema used for importing data as soon as possible to reduce spike in disk usage
- while materialized views should be refreshed synchronously as part of the process, precomputed downloads can be generated asynchronously after the upload is finished
- finally, there was a stupid mistake because of which chart attributes would get lost after update

### DIFF
--- a/app/models/api/v3/chart_ind.rb
+++ b/app/models/api/v3/chart_ind.rb
@@ -28,7 +28,7 @@ module Api
 
       def self.yellow_foreign_keys
         [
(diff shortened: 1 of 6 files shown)


agnessa:
This branch is already deployed on sandbox, as it fixes an important bug related to database uploads + profile pages. It's important that this fix remains active on sandbox while the database uploader is being tested, so if any changes from develop need to be deployed it is important to merge this first.

PR #706, opened 2019-02-18
Feature/use axios

Bluesmile82:
This PR uses axios to add promise cancellation also fixes some bugs in the dashboard selection.

The sagas have been refactored to include the logic that previously was in the actions. There are splitted into saga and fetch.saga. The latter are in chage of doing the fetchs to the API with axios and cancelling if needed.


frontend/scripts/react-components/dashboard-element/dashboard-element.actions.js
```diff
@@ -153,66 +101,8 @@ export const setDashboardPanelLoadingItems = loadingItems => ({
...
-    optionsType === 'sources' &&
-    dashboardElement.sourcesPanel.activeItem === null &&
-    dashboardElement.sourcesPanel.activeTab === null
-  ) {
-    optionsType = 'countries';
-  }
-  const filters = {
-    ...getDashboardPanelParams(dashboardElement, optionsType),
-    node_types_ids: undefined
+export const getDashboardPanelSearchResults = (query, bla) =>
```
sorodrigo: this is probably not being used

frontend/scripts/react-components/dashboard-element/dashboard-element.fetch.saga.js
```diff
@@ -0,0 +1,166 @@
...
+        data: data.data,
+        meta: data.meta,
+        loading: false
+      }
+    });
+  } catch (e) {
+    console.error('Error', e);
+  } finally {
+    if (yield cancelled()) {
+      console.error('Cancelled', tab);
```
sorodrigo: Let's either remove or hide in prod

frontend/scripts/react-components/dashboard-element/dashboard-indicators/dashboard-indicators.container.jsx
```diff
@@ -32,17 +30,12 @@ const mapStateToProps = state => ({
...
     indicators: PropTypes.array,
-    getData: PropTypes.func.isRequired,
     activeIndicatorsList: PropTypes.array,
     dynamicSentenceParts: PropTypes.array,
     setActiveId: PropTypes.func.isRequired,
     removeActiveId: PropTypes.func.isRequired
   };
 
-  componentDidMount() {
-    this.props.getData('attributes');
```
sorodrigo: This was needed after all, the indicators list is not being loaded now. Let's discuss about the best way to fix this.

PR #707, opened 2019-02-19
Feature/dashboard testing

Bluesmile82:
Tests dashboard sagas and dashboards fetch saga files.

We need to export the saga functions and keep them on one level to be able to test them.
Fetching action its not tested itself but cancel is.
We are using this methodology to test only the final actions dispatched by the sagas: https://dev.to/phil/the-best-way-to-test-redux-sagas-4hib

The recordSaga syntax has been updated due to changes in the redux-saga API
It will be interesting to research more in the sagaMonitor object created by runSaga:
See the runSaga definition in https://redux-saga.js.org/docs/api/

In the fetch saga file, only the cancellation is tested but It can surely be improved to prevent a step by step testing. Any ideas are appreciated.

### DIFF
--- a/frontend/scripts/tests/dashboard-element/saga.test.js
+++ b/frontend/scripts/tests/dashboard-element/saga.test.js
@@ -1,10 +1,3 @@
-import { fork, select } from 'redux-saga/effects';
-import {
-  getDashboardPanelSectionTabs,
-  getDashboardPanelData,
-  fetchDashboardPanelSearchResults,
-  getMoreDashboardPanelData
-} from 'react-components/dashboard-element/dashboard-element.fetch.saga';
 import {
   fetchDashboardPanelInitialData,
   getSearchResults,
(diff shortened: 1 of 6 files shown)


Bluesmile82:
Updated! Check description and changes

frontend/scripts/tests/dashboard-element/saga.test.js
```diff
@@ -0,0 +1,487 @@
...
+        meta: null,
+        loading: true
+      },
+      type: 'DASHBOARD_ELEMENT__SET_PANEL_DATA'
+    });
+  });
+});
+
+describe('getSearchResults', () => {
+  const searchAction = {
```
sorodrigo: we should use action creator

frontend/scripts/tests/dashboard-element/saga.test.js
```diff
@@ -0,0 +1,487 @@
...
+      countriesPanel: {
+        activeTab: {
+          id: 2
+        },
+        page: 2
+      }
+    }
+  };
+
+  const clearAction = {
```
sorodrigo: same

PR #708, opened 2019-02-19
fix: showing incorrect results + clearing wrong tab data

sorodrigo:
This PR address a couple of issues I found on the sources panel:
- Search in sources should show only countries when no country is selected. If you selected a country, then deselect it and search again, the results were not showing countries but jurisdictions.
- When changing tab using the search, the page number wasn't being reset.
- When changing tabs the data being reset was that of the new tab, instead of the previous.

### DIFF
--- a/frontend/scripts/react-components/dashboard-element/dashboard-panel/dashboard-panel.container.jsx
+++ b/frontend/scripts/react-components/dashboard-element/dashboard-panel/dashboard-panel.container.jsx
@@ -88,6 +88,11 @@ class DashboardPanelContainer extends React.PureComponent {
     }
   ];
 
+  static propTypes = {
+    companiesPanel: PropTypes.object,
+    countryNames: PropTypes.object
+  };
+
   static addCountryNameToSearchResults(panel, countryNames) {
     const searchResults = panel.searchResults.map(item => ({
       ...item,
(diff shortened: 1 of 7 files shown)


frontend/scripts/react-components/dashboard-element/dashboard-panel/dashboard-panel.container.jsx
```diff
@@ -88,6 +88,11 @@ class DashboardPanelContainer extends React.PureComponent {
     }
   ];
 
+  static propTypes = {
```
Bluesmile82: Should we start using the es7 syntax always?

frontend/scripts/react-components/shared/tags-group/tags-group.component.jsx
```diff
@@ -21,10 +21,7 @@ function TagsGroup(props) {
           >
             {part.value}
             {clearItem && (
-              <button
-                onClick={() => console.log('click', part) || clearItem(part)}
-                className="tags-group-item-remove-cross"
-              >
+              <button onClick={() => clearItem(part)} className="tags-group-item-remove-cross">
```
Bluesmile82: Thanks

frontend/scripts/react-components/dashboard-element/dashboard-panel/dashboard-panel.container.jsx
sorodrigo: I'm trying to use it when we have class properties

PR #710, opened 2019-02-20
Use levenshtein search algorithm in tool search

sorodrigo:
This PR changes the algorithm in the tool search. We were using `indexOf` now, we're using _Levenshtein_.

Adds a feature flag called `ENABLE_LEGACY_TOOL_SEARCH` to revert to original algo please test that it works.


sorodrigo:
…Oh, do a little 🕺 , make a little ❤️ , get ⬇️  🌃 , get ⬇️  🌃 .

PR #711, opened 2019-02-21
Actually apply the per_page param for index endpoints

agnessa:


aabdaab:
checking ☝️

aabdaab:
Everything's working fine 👌 i'm wondering if we should be relying on a specific object being in the array in the specs, what if that requirement changes. I think it'd be better to check the count for `per_page` query param. we could also move it to its own variable and not hardcode the result. i.e.:
```
let(:per_page) { 2 }
it 'accepts per_page' do
  get :index, params: {countries_ids: api_v3_brazil.id, per_page: per_page}
  expect(assigns(:collection).length).to eq(per_page)
end
```

agnessa:
>  i'm wondering if we should be relying on a specific object being in the array in the specs, what if that requirement changes. I think it'd be better to check the count for `per_page` query param. we could also move it to its own variable and not hardcode the result. 
done!

PR #712, opened 2019-02-21
[BE] Fix selecting multiple companies

aabdaab:
* Fix filtering in the dashboards/chart_data endpoint - use OR operator instead of AND for filtering companies_ids
* Fix filtering in the dashboards/attributes endpoint - use OR operator instead of AND for filtering companies_ids
* Solution: replace ARR1 @> ARR2 operator that checks for EVERY element from ARR2 in ARR1 with && operator checking if the they have ANY elements in common

PT task:
https://www.pivotaltracker.com/n/projects/1637931/stories/164108523
And also:
https://www.pivotaltracker.com/n/projects/1637931/stories/164108677
https://www.pivotaltracker.com/n/projects/1637931/stories/164109281
As the `nodes_ids` in `RetrieveAttributes` service include all `companies_ids`, `destinations_ids` and `sources_ids`.

Testing:
Check `/api/v3/dashboards/chart_data?attribute_id=31&companies_ids=32228,29138` endpoint which previously didn't return any results

### DIFF
--- a/spec/controllers/api/v3/dashboards/chart_data_controller_spec.rb
+++ b/spec/controllers/api/v3/dashboards/chart_data_controller_spec.rb
@@ -39,10 +39,62 @@ RSpec.describe Api::V3::Dashboards::ChartDataController, type: :controller do
       expect(data.map { |e| e['x'] }).to eq([2015])
     end
 
-    it 'returns values for multiple matched companies' do
+    it 'returns values for a single company - filtering by exporter node' do
+      get :index, params: {
+        attribute_id: api_v3_volume.readonly_attribute.id,
+        companies_ids: [api_v3_exporter1_node.id].join(',')
+      }
+
+      json = JSON.parse(response.body)
+      data = json['data']
+      expect(data.map { |e| e['x'] }).to eq([2015])
+      expect(data.map { |e| e['y1']}).to eq([
+        flow1_volume.value +
+        flow2_volume.value +
+        flow3_volume.value +
+        flow5_volume.value
+      ])
+    end
+
+    it 'returns values for multiple companies - filtering by importer and exporter node' do
       get :index, params: {
         attribute_id: api_v3_volume.readonly_attribute.id,
         companies_ids: [api_v3_exporter1_node.id, api_v3_importer1_node.id].join(',')
+      }
+
+      json = JSON.parse(response.body)
+      data = json['data']
+      expect(data.map { |e| e['x'] }).to eq([2015])
+      expect(data.map { |e| e['y1']}).to eq([
+        flow1_volume.value +
+        flow2_volume.value +
+        flow3_volume.value +
+        flow4_volume.value +
+        flow5_volume.value
+      ])
+    end
+
+    it 'returns values for multiple companies - filtering by 2 exporters' do
+      get :index, params: {
+        attribute_id: api_v3_volume.readonly_attribute.id,
+        companies_ids: [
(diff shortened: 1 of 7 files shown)


aabdaab:
@agnessa ☝️ ready again, added specs for `chart_data` endpoint and one for `destinations` by accident but I decided to leave it ;) 
Please, have a look as I'm still uncertain about the business logic behind some of the cases.

spec/support/contexts/api/v3/brazil/brazil_flows.rb
```diff
@@ -13,7 +13,8 @@
         api_v3_port1_node,
         api_v3_exporter1_node,
         api_v3_importer1_node,
-        api_v3_country_of_destination1_node
+        api_v3_country_of_destination1_node,
+        api_v3_other_country_of_destination_node
```
agnessa: We cannot do that - there can be only one country of destination in a single flow. In case of Brazil soy, each flow needs to have a path made of 8 nodes and the last of those is the country of destination. What you need here is to create another flow for your test; you could try adding it in the shared context, but it might break other specs that load it, so if that happens add it in your spec file, e.g. 

```
  let!(:api_v3_flow3) do
    FactoryBot.create(
      :api_v3_flow,
      context: api_v3_context,
      path: [
        api_v3_biome_node,
        api_v3_state_node,
        api_v3_municipality_node,
        api_v3_logistics_hub_node,
        api_v3_port1_node,
        api_v3_exporter1_node,
        api_v3_importer1_node,
        api_v3_other_country_of_destination_node
      ].map(&:id),
      year: 2015
    )
  end
```

You need to do a similar thing in order to be able to test filtering by 2 exporters - you need to create an extra flow with a different exporter node, e.g.:

```
  let!(:api_v3_flow4) do
    FactoryBot.create(
      :api_v3_flow,
      context: api_v3_context,
      path: [
        api_v3_biome_node,
        api_v3_state_node,
        api_v3_municipality_node,
        api_v3_logistics_hub_node,
        api_v3_port1_node,
        api_v3_other_exporter_node,
        api_v3_importer1_node,
        api_v3_country_of_destination1_node
      ].map(&:id),
      year: 2015
    )
  end
```

spec/controllers/api/v3/dashboards/chart_data_controller_spec.rb
```diff
@@ -38,5 +38,30 @@
       data = json['data']
       expect(data.map { |e| e['x'] }).to eq([2015])
     end
+
+    it 'returns values for multiple matched companies' do
```
agnessa: This test is fine, but it tests what used to work previously, which is filtering by importer and exporter node; what we should add is another test which tests filtering by 2 exporters or 2 importers, as this is the functionality that's new.

aabdaab:
Thanks a lot for these explanations, now it is clear for me. I'm wondering if it would be beneficial to start out with some domain documentation, for potential new developers' reference but also for us? Check out the updated specs ☝️

agnessa:
👌 I added a spec to test for single exporter vs multiple exporters considering the returned values as well - I think this kind of test is going to be useful for the other multiple filters. Agree on the docs. We're now working on improving the documentation, I think a domain section would be great. We're obviously a bit stretched for time this quarter, so not sure when that might be ready. I started a table of contents [here](https://github.com/Vizzuality/trase/issues/717), would be good to have your inputs - or maybe a list of specific questions you need answered.

PR #714, opened 2019-02-22
Fix/grid list pagination

sorodrigo:
This PR fixes some issues in the grid-list and in the add more data action.
- Don't fetch data when scrolling backwards
- Don't increase the page number after reaching the end.


Bluesmile82:
It seems like the page number is not consistent. Sometimes it starts with page 2 or 3. We may want to fix this to prevent loading more data when we reach the top. Also is there any way to prevent the +1 / -1 page changes when we reach the last page. Maybe this is related

PR #715, opened 2019-02-24
Fix: prevent dashboards unnecessary fetch

sorodrigo:
So I finally tackled the annoying situation where changing panels would cause data reload even if nothing had actually changed. The solution is pretty cool, I'm using `redux-saga's take()` effect to calculate the changes between panel changes 😏 .

Besides that, I discovered another bug on the tabs selection:

#### Current Behavior
1. Fetch new tabs according to selection change.
2. If a selected tab existed prior to tabs fetch, use it as default, otherwise use `tabs[0]`.

Problem? We weren't checking if the selected tab also existed in the new tabs, leaving an invalid selection as default.

#### New Behavior
1. Fetch new tabs according to selection change.
2. If a selected tab existed prior to tabs fetch and exists in new tabs, use it as default, otherwise use `tabs[0]`.

Another issue that is fixed here:
When selecting and deselectin an item the page would reload.

**TODO**
- [x] Add regression test for tabs default value.
- [x] Add test for `fetchDataOnPanelChange` saga.

### DIFF
--- a/frontend/scripts/react-components/dashboard-element/dashboard-element.saga.js
+++ b/frontend/scripts/react-components/dashboard-element/dashboard-element.saga.js
@@ -140,15 +140,19 @@ function* fetchDataOnItemChange() {
 }
 
 /**
- * Listens to DASHBOARD_ELEMENT__CLEAR_PANEL and fetches the necessary data after a filter clear
+ * Listens to DASHBOARD_ELEMENT__CLEAR_PANEL and fetches the necessary data after a filter clear.
+ * On sources and companies we don't need to call getDashboardPanelData because getDashboardPanelSectionTabs
+ * dispatches an action that trigger it.
  */
 export function* onFilterClear() {
   const { dashboardElement } = yield select();
   if (dashboardElement.activePanelId === 'sources') {
     yield fork(getDashboardPanelData, dashboardElement, 'countries');
     if (dashboardElement.countriesPanel.activeItem !== null) {
-      yield fork(getDashboardPanelData, dashboardElement, 'sources');
+      yield fork(getDashboardPanelSectionTabs, dashboardElement, dashboardElement.activePanelId);
     }
+  } else if (dashboardElement.activePanelId === 'companies') {
+    yield fork(getDashboardPanelSectionTabs, dashboardElement, dashboardElement.activePanelId);
   } else {
     yield fork(getDashboardPanelData, dashboardElement, dashboardElement.activePanelId);
   }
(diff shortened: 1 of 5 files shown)


sorodrigo:
This will need to change once we support the multiple selection of items, so let's wait until then to add tests and merge.

frontend/scripts/react-components/dashboard-element/dashboard-element.saga.js
```diff
@@ -36,7 +36,44 @@ export function* fetchDashboardPanelInitialData(action) {
...
-  yield takeLatest(DASHBOARD_ELEMENT__SET_ACTIVE_PANEL, fetchDashboardPanelInitialData);
+  let loaded = [];
+  let activePanelState = null;
+  const hasChanged = panel => {
+    if (!activePanelState) return false;
+    const changes = [];
+    if (panel.sourcesPanel.activeItem !== activePanelState.sourcesPanel.activeItem) {
+      changes.push('sources');
+    }
+    if (panel.countriesPanel.activeItem !== activePanelState.countriesPanel.activeItem) {
```
Bluesmile82: Maybe we could write some of this together like:

```
const fullPanelName = `${panelName]Panel`;

if (panel[fullPanelName) !== activePanelState[pafullPanelNameelName].activeItem) {
  changes.push(panelName);
}
```

frontend/scripts/react-components/dashboard-element/dashboard-element.saga.js
```diff
@@ -36,7 +36,44 @@ export function* fetchDashboardPanelInitialData(action) {
...
+  };
+
+  while (true) {
+    const activePanel = yield take(DASHBOARD_ELEMENT__SET_ACTIVE_PANEL);
+    const newPanelState = yield select(state => state.dashboardElement);
+    const changes = hasChanged(newPanelState);
+    if (changes) {
+      loaded = changes;
+    }
+    if (!activePanelState || !loaded.includes(activePanel.payload.activePanelId)) {
```
Bluesmile82: We could use a const for activePanel.payload.activePanel

frontend/scripts/react-components/dashboard-element/dashboard-element.saga.js
```diff
@@ -85,7 +126,10 @@ export function* onItemChange(action) {
   if (panel === 'countries') {
     yield fork(getDashboardPanelSectionTabs, dashboardElement, panel);
   } else if (items && !activeItemExists) {
-    yield fork(getDashboardPanelData, dashboardElement, panel);
+    // only reload list if clearing item from a different panel than the one that is active
+    // if (!activeItem && panel !== dashboardElement.activePanelId) {
```
sorodrigo: I just realised this is commented, can't understand why this line is here. If we uncomment it leaving the if statement, it also works. I wonder if this is necessary? Thoughts @Bluesmile82 ?

sorodrigo:
I discovered yet another bug 😪 , the tabs weren't recalculated after clearing an option. I added a fix and test. I'm a bit concerned with having to work on stuff that might change later. Oh well, it's ready.

PR #716, opened 2019-02-25
Feature/dashbords panels multiple selection

Bluesmile82:
This PR makes possible the multiple selection of regions, importing countries, companies and commodities.

Note: The dynamic sentence is not updated yet

Extra: [directory-named-webpack-plugin](https://www.npmjs.com/package/directory-named-webpack-plugin) is installed and used in webpack so we can use the import as: `'react-components/whatever'` instead of: `'react-components/whatever/whatever.container'`
We agreed that this new index file will be the container exporting the actions and reducers when needed. One good reason for this is consistency with other projects on the company, also to avoid an extra file.

image


frontend/scripts/react-components/dashboard-element/dashboard-element.reducer.js
```diff
@@ -34,39 +38,44 @@ const initialState = {
...
     page: 1,
     searchResults: [],
     loadingItems: false,
-    activeItem: null,
+    activeItems: {},
     activeTab: null
   }
 };
 
+const activeItems = (currentItems, newItem) =>
```
sorodrigo: Let's rename this to something more significant, `toggleSelectedItem`?

frontend/scripts/tests/dashboard-element/reducer.test.js
```diff
@@ -694,7 +696,101 @@ test(DASHBOARD_ELEMENT__SET_ACTIVE_ITEM_WITH_SEARCH, () => {
...
+      ...state,
+      activeIndicatorsList: [],
+      companiesPanel: {
+        ...state.companiesPanel,
+        activeItems: { [someItem.id]: someItem }
+      }
+    });
+  });
+
+  it('sets active item in the countries panel', () => {
```
sorodrigo: is this test correct? I think countries doesn't allow more than 1 activeItem

frontend/scripts/tests/dashboard-element/saga.test.js
```diff
@@ -326,7 +326,7 @@ describe('onFilterClear', () => {
       },
       countriesPanel: {
         ...baseState.dashboardElement.countriesPanel,
-        activeItem: null
+        activeItems: {}
```
sorodrigo: is this correct?

frontend/.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
sorodrigo: please remove this file https://help.github.com/en/articles/ignoring-files

PR #718, opened 2019-02-26
Fix/automate db documentation

agnessa:
This automates generating the db documentation using Travis CI deploy pages feature. On the master branch it builds the docs using `rake db:doc:html` and then pushes the outputs to the `gh-pages` branch. The published schema documentation is at https://vizzuality.github.io/trase/all_tables/.

Note: before I set the branch to master, I tested it on the current branch, so you can check out the automated commits in https://github.com/Vizzuality/trase/tree/gh-pages

PS: Travis also supports deploying to surge: https://docs.travis-ci.com/user/deployment/surge/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,8 @@ matrix:
         - postgresql
       addons:
         postgresql: 10
+      before_install:
+        - sudo apt-get install -y graphviz
       before_script:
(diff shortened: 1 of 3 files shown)


aabdaab:
LGTM, we can just apply the Codacy suggestion, then feel free to merge

PR #720, opened 2019-02-28
Fix/dashboard dynamic sentence

Bluesmile82:
This PR branches from: `feature/dashbords-panels-multiple-selection`
First commit is: Use several items in dynamic sentence, split importing and exporting … 1780acd

Fix the dynamic sentence in the dashboard modal and view:

- Add a dropdown component based on downshift
- Use it to make possible to view and clear options from the dropdown
- Use it with different props to view the options in the dashboard view
- The dropdown in the modal opens upwards due to lack of space downwards. It was also blocking the continue button

image

image


sorodrigo:
In the meantime for the dashboard dropdown, we can make it so that the styles match the Panel dropdown and also add a `title` html property to the text so that it can be read on hover.

Bluesmile82:
Comments addressed! Merging

PR #723, opened 2019-03-04
Add tables to support context specific, country specific and commodit…

aabdaab:
…y specific properties for inds, quants and quals

This PR adds 9 tables (3 per context, 3 per country, 3 per commodity) to support specific properties (tooltips) respectively for context, country and commodity for inds, quals and quants.
I also added three materialized views which combine all attributes (inds, quals and quants) in one view.
The main change resides in `sustainability_table` service.

I have not worked with materialized views before, so double-check if it makes sense 🙏

### DIFF
--- a/app/services/api/v3/actors/sustainability_table.rb
+++ b/app/services/api/v3/actors/sustainability_table.rb
@@ -11,6 +11,7 @@ module Api
           @context = context
           @node = node
           @year = year
+
           # Assumption: Volume is a special quant which always exists
           @volume_attribute = Dictionary::Quant.instance.get('Volume')
           raise 'Quant Volume not found' unless @volume_attribute.present?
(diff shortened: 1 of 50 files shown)


aabdaab:
^ Also, working on specs, will add them shortly.

app/services/api/v3/actors/sustainability_table.rb
```diff
@@ -135,6 +136,26 @@ def totals_row(group_totals_hash)
               end
           }
         end
+
+        def get_tooltip(ro_chart_attribute)
```
agnessa: I guess we'll need the same logic for some of the place profile charts, so I'd consider moving this function to its own service / module under the `Api::V3::Profiles` namespace for example (might also make it easier to write specs)

aabdaab: good idea about extracting this. I added a separate module in the next commit, will add specs to it in another one

aabdaab:
Ok, ready again for review. Since this feature has its own feature branch I think we can leave the MVs changes for the next PRs (and since they've got their own PT tasks).

PR #724, opened 2019-03-06
Context specific tooltips

aabdaab:
This is a joint branch for implementing specific tooltips texts regarding context/country/commodity aspects.


agnessa:
Would be great to improve test coverage here.

aabdaab:
Ready for review ☝️ 
I added missing specs and fixed most of the code climate issues (there are two left but i'd say irrelevant, especially the complain about `V3` module name).

PR #727, opened 2019-03-07
Data uploader changes

aabdaab:
* Include new tables in the import script
* Include new MVs in the import script
* Add views to Data Uploader
* Add data validations in the models

**Doubts**
Is it correct that I'm inheriting from **YellowTable** in the new models? I changed it as I tried to follow what was in the import/tables module (with the Ind/Quant/QualProperty being in the yellow_tables) but I'm not entirely sure what it means.

### DIFF
--- a/app/admin/api/v3/chart_attribute.rb
+++ b/app/admin/api/v3/chart_attribute.rb
@@ -32,7 +32,7 @@ ActiveAdmin.register Api::V3::ChartAttribute, as: 'ChartAttribute' do
             required: true,
             collection: Api::V3::Chart.select_options
       input :identifier, as: :string,
-            hint: object.class.column_comment('position')
+                         hint: object.class.column_comment('position')
       input :position, hint: object.class.column_comment('position')
       input :display_name, as: :string,
                            hint: object.class.column_comment('display_name')
--- a/app/admin/api/v3/dashboard_template.rb
+++ b/app/admin/api/v3/dashboard_template.rb
@@ -2,8 +2,8 @@ ActiveAdmin.register Api::V3::DashboardTemplate, as: 'DashboardTemplate' do
   menu parent: 'Dashboards'
 
   permit_params :title, :description, :image, :category, commodity_ids: [],
-                company_ids: [], country_ids: [], destination_ids: [],
-                source_ids: []
+                                                         company_ids: [], country_ids: [], destination_ids: [],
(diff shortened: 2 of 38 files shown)


agnessa:
As discussed, let's add `NOT NULL` to all foreign keys and `tooltip_text` columns in the new tables, and a presence validation for `tooltip_text` in the AR model.

Another small thing is it would be great to inform the user somehow via the CMS about the precedence of tooltips. E.g. if they could see a short info text on the different CMS pages e.g. when you're on country tooltips "Country-specific tooltips are the second-most specific tooltips that can be defined after context-specific tooltips (link); in absence of a country-specific tooltip, a commodity-specific tooltip (link) will be used."; we could also order links in the menu by precedence.

It would be great if we could have a submenu or subnav for the tooltip related links, as that dropdown menu became quite long - but not sure how easy that is, so that's less important if no standard approach exists.

In the CMS we need to consider clearing the cache. This most importantly handles the Varnish cache, which is hard to test locally, but it should be safe to copy that logic from the respective ind/qual/quant_properties pages (`after_action :clear_cache`).

You'll notice in the `clear_cache` method that in addition to clearing Varnish cache this also resets some singleton dictionaries. I am a bit worried about those, because I don't remember anymore where they are used, and they currently load the generic tooltip only. It would be worth having a look at those dictionaries as part of the "API changes" task and see if they could be removed entirely.

agnessa:
One more thing related to specs: notice that now that the `refresh_dependents` callbacks are included, materialized views are refreshed twice per test, once from callback and once from `refresh_dependent_materialized_views`. The way we deal with this is to disable `refresh_dependents` for specs, but it depends for now on maintaining a list of the triggering classes in `rails_helper.rb`, so could you please add the new models to that list - that should prevent the double refresh in specs.

aabdaab:
It is ready again for review;
I added:
- Descriptions for CMS pages; added them directly in the index block, I wanted to avoid creating partials for that purpose
- I extracted tooltips from the General tab and moved them to dedicated Tooltips tab and I ordered them by precedence 
- Added `clear_cache` method to the admin pages (it seems like we don't need to reset the Dictionaries, I haven't found any occurrence in the code that would make use of tooltip_text from a dictionary. It's widely used to get a specific volume attribute across the platform.
- Disabled refreshing dependents for specs
- I also added `on delete cascade`, `on update cascade` to new tables

PR #728, opened 2019-03-08
API changes

aabdaab:
Get correct tooltip in the rest of the places where we retrieve `tooltip_text`; `IndicatorsTable` and `TrajectoryDeforestationPlot `.
PT task: https://www.pivotaltracker.com/n/projects/1637931/stories/163422246


agnessa:
Great! We need to check if there are any other places where generic tooltips might be showing. I went through the places where I'm aware some tooltips are present and it seems that:
- sankey - recolor by / resize by: this is probably ok, because there are specific tooltips for that
- sankey - map: I think the info buttons next to layers show generic tooltips
- place profiles: info buttons next to "soy land" and "soy production" probably show generic tooltips - these are returned by the basic_attributes endpoint
- actor profiles: similarly, info buttons next to "zero deforestation commitment" and "forest 500 score" (basic_attributes endpoint)
- nothing in data downloads page
- dashboards: possibly generic tooltips returned in the attributes endpoint, but perhaps this can be left out from this PR, because that endpoint will need to be changed in other ways as well

Hope this is it! Sorry, it's hard to remember all the places where things are used 😅

aabdaab:
About the tooltips on the layers next to the sankey-map; so the tooltips are actually defined in `Api::V3::ContextualLayer`. It is related with specific context so following our logic it is the most specific tooltip for the layer. Do we need less specific tooltips for the layers then, like in the case of quant/qual/ind?

agnessa:
You're right about contextual layers - those are probably fine, but the ones above them ("map attributes", e.g. "agricultural indicators") probably not.

aabdaab:
huh, these are interesting cause these are actually hardcoded in the `tooltips.json` file

agnessa:
I think they come from the map_layers endpoint; what the `tooltips.json` does I have no idea 😮

aabdaab:
On `profile_place` the tooltips next to Soy Land and Soy Production (https://trase.earth/profile-place?nodeId=9020&contextId=1&year=2017) also come from the `tooltips.json` file

aabdaab:
@agnessa I added tooltips logic into `map_layers` endpoint. I suggest leaving the `tooltips.json` issue for another PR, as a first step we can ask the Trase team how they feel about these certain tooltips that don't reside in the DB.
As for the `attributes` endpoint, the change as you said isn't so straightforward and we'd need additional info there

PR #730, opened 2019-03-13
Feature/endpoint returning dashboards chart urls

agnessa:
New endpoint that dynamically calculates the types of charts to be rendered for selected parameters and returns the chart types and data urls.

This follows the spec in ["Dashboards filters: Visualizations to be used for each filter selection"](https://docs.google.com/document/d/1jBMhDs4_51cW4xC3fSH2AboGj934OPh-DEMFPFyY7b0/edit?usp=sharing). Adds a new endpoint which accepts the currently selected filters and parameters, and returns a list of chart types and urls to fetch data for the charts. The data endpoints are not finished, just the routes were added here and implementation will follow in further PRs.

Swagger docs were updated.

### DIFF
--- a/app/controllers/api/v3/dashboards/parametrised_charts_controller.rb
+++ b/app/controllers/api/v3/dashboards/parametrised_charts_controller.rb
@@ -1,0 +1,23 @@
+module Api
+  module V3
+    module Dashboards
+      class ParametrisedChartsController < ApiController
+        include FilterParams
+        skip_before_action :load_context
+
+        def index
+          ensure_required_param_present(:country_id)
+          ensure_required_param_present(:commodity_id)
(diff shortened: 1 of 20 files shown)


aabdaab:
we could think about adding better error handling; for now when record for passed params is not found the page crashes - we could add meaningful error description like in the case of not passing required params:
image

agnessa:
Added error payload to 404 responses. I had to amend the profile specs, as they were testing for an exception, and now they need to test for the response status. I checked that the profile pages still behave the same after a not found exception.

agnessa:
It would be great to re-review this one, as there are more chart-related PRs coming. Thanks!

aabdaab:
^ checking

PR #735, opened 2019-03-18
Feature/overview chart data endpoints

agnessa:
This adds 4 "overview" chart endpoints, which return data for different types of charts, one per each of the 4 broad types of dashboards parameters selections:
1. when single year selected and no non-continuous indicator selected - dynamic sentence
2. when single year selected and a non-continous indicator selected - donut chart
3. when multiple years selected and no non-continuous indicator selected - vertical bar chart
4. when multiple years selected and a non-continuous indicator selected - vertical stacked bar chart

This corresponds to the [spec](https://docs.google.com/document/d/1jBMhDs4_51cW4xC3fSH2AboGj934OPh-DEMFPFyY7b0/edit?usp=sharing). Responses described in the swagger file.

### DIFF
--- a/app/controllers/api/v3/dashboards/charts/multi_year_ncont_overview_controller.rb
+++ b/app/controllers/api/v3/dashboards/charts/multi_year_ncont_overview_controller.rb
@@ -1,0 +1,25 @@
+module Api
+  module V3
+    module Dashboards
+      module Charts
+        class MultiYearNcontOverviewController < ApiController
+          include FilterParams
+          skip_before_action :load_context
+
+          def index
+            ensure_required_param_present(:country_id)
+            ensure_required_param_present(:commodity_id)
+            ensure_required_param_present(:start_year)
+            ensure_required_param_present(:end_year)
+            ensure_required_param_present(:cont_attribute_id)
+            ensure_required_param_present(:ncont_attribute_id)
+
+            render json: MultiYearNcontOverview.new(
+              ChartParameters.new(chart_params)
+            ).call
+          end
+        end
+      end
+    end
+  end
+end
--- a/app/controllers/api/v3/dashboards/charts/multi_year_no_ncont_overview_controller.rb
+++ b/app/controllers/api/v3/dashboards/charts/multi_year_no_ncont_overview_controller.rb
@@ -1,0 +1,24 @@
+module Api
+  module V3
+    module Dashboards
+      module Charts
+        class MultiYearNoNcontOverviewController < ApiController
+          include FilterParams
+          skip_before_action :load_context
+
+          def index
+            ensure_required_param_present(:country_id)
+            ensure_required_param_present(:commodity_id)
+            ensure_required_param_present(:start_year)
+            ensure_required_param_present(:end_year)
+            ensure_required_param_present(:cont_attribute_id)
+
+            render json: MultiYearNoNcontOverview.new(
+              ChartParameters.new(chart_params)
+            ).call
+          end
+        end
+      end
+    end
+  end
+end
--- a/app/controllers/api/v3/dashboards/charts/single_year_ncont_overview_controller.rb
+++ b/app/controllers/api/v3/dashboards/charts/single_year_ncont_overview_controller.rb
@@ -1,0 +1,24 @@
+module Api
+  module V3
+    module Dashboards
+      module Charts
+        class SingleYearNcontOverviewController < ApiController
+          include FilterParams
+          skip_before_action :load_context
+
+          def index
+            ensure_required_param_present(:country_id)
+            ensure_required_param_present(:commodity_id)
+            ensure_required_param_present(:start_year)
+            ensure_required_param_present(:cont_attribute_id)
+            ensure_required_param_present(:ncont_attribute_id)
+
+            render json: SingleYearNcontOverview.new(
+              ChartParameters.new(chart_params)
+            ).call
+          end
+        end
+      end
+    end
+  end
+end
--- a/app/controllers/api/v3/dashboards/charts/single_year_no_ncont_overview_controller.rb
+++ b/app/controllers/api/v3/dashboards/charts/single_year_no_ncont_overview_controller.rb
@@ -1,0 +1,23 @@
+module Api
+  module V3
+    module Dashboards
+      module Charts
+        class SingleYearNoNcontOverviewController < ApiController
+          include FilterParams
+          skip_before_action :load_context
+
+          def index
+            ensure_required_param_present(:country_id)
+            ensure_required_param_present(:commodity_id)
+            ensure_required_param_present(:start_year)
+            ensure_required_param_present(:cont_attribute_id)
+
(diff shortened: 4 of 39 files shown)


agnessa:
@sorodrigo I would like to double check about the chart data and meta responses. I think I'm more unsure about the meta. Can you take the look at the swagger - it now includes response formats for the 4 "overview" charts, i.e. dynamic sentence, donut chart, bar chart and stacked bar chart (you can find them in the "charts" section). I use the same response structure that we had previously, based on our conversations back in October, but there were always some aspects of it I wasn't sure about, so would be good to double check.

Some of my doubts:
- the `format` property is always empty (because I don't know what to set it to); I have a feeling it has to do with the `prefix` / `suffix` situation, but I'm not sure what the correct value of that would be? For now we're only using `suffix` for units (`prefix` is always empty), because even if it is a "prefix"-type unit, at this stage we have no way of knowing that; but in the future we might, so maybe worth having the format set correctly.
- I think I understand `meta->xAxis` and `meta->yAxis`, but double checking: this is what goes on axis labels. In case there's an indicator for either axis, this will give its name (using `label`) and unit (using `suffix`). In case it's a year (for x axis) the `label` just says 'Year' and the rest of the attributes are not used.
- I'm not sure why the `type` (`category`,  `date`, `number`) is not part of `meta->xAxis, yAxis`; instead it's part of `meta->x, yn`, which doesn't make sense to me that e.g. in case of the stacked bar chart, where we have multiple `yn`, we'd need to repeat the same `type` property for each of them (don't see how they could have different types), but maybe it makes sense?
- I don't understand `meta->x` at all; in all cases it seems to me that it's going to be equivalent to the information in `meta->xAxis` (except for the `type`, as discussed above)?
- I understand `meta->yn` as values for the legend in the stacked bar chart. For now these are mostly values of the non-continuous indicator, but in the upcoming "break down by node type" charts this could be the name of the node as well.
- I'm not sure what the value of `meta->yn->type` should be in the stacked bar chart. Should it be the type of value plotted on the y axis (which is going to be always numeric, e.g. trade volume) or should it reflect the attribute / node by which values are aggregated (which is going to be always categorical, e.g. biome or value of forest 500).
- What I don't entirely understand is why `meta->x, yn` has a different format than `meta->xAxis, yAxis`, i.e. in the first case it contains the `tooltip` object which packages the `suffix`, `prefix` and `format` properties (leaving `label` outside), whereas in the latter all 4 properties are in the same level of the json. Not much difference really, but I'm not sure there's a good reason for it, so just wondering if it would be worth simplifying and formatting them in the same way.

sorodrigo:
>I'm not sure why the type (category, date, number) is not part of meta->xAxis, yAxis; instead it's part of meta->x, yn, which doesn't make sense to me that e.g. in case of the stacked bar chart, where we have multiple yn, we'd need to repeat the same type property for each of them (don't see how they could have different types), but maybe it makes sense?

You're right, in practice we're using it for the `xAxis / yAxis` so it makes more sense to have them there.

> I don't understand meta->x at all; in all cases it seems to me that it's going to be equivalent to the information in meta->xAxis (except for the type, as discussed above)?

Again you're right, we're not actually using it 🤦‍♂️ .

This is how we're parsing meta for the axes:
```js
xAxis: defaultConfig.xAxis && {
  ...defaultConfig.xAxis,
  type: meta.x && meta.x.type
},
yAxis: defaultConfig.yAxis && {
  ...defaultConfig.yAxis,
  type: meta.y && meta.x.type
}
```

sorodrigo:
> I'm not sure what the value of meta->yn->type should be in the stacked bar chart. Should it be the type of value plotted on the y axis (which is going to be always numeric, e.g. trade volume) or should it reflect the attribute / node by which values are aggregated (which is going to be always categorical, e.g. biome or value of forest 500).

So if we move the type to the yAxis key, this gets solved. For a stacked bar, we need to provide a `stackId` on which we'll group the bars. This for now can be a generic hardcoded key such as `bar`, because we're building charts that are solely stacked bars. However, if we wished to build a chart that had stacked bars alongside normal bars, we would need different `stackId` values. (I'll attach an example chart)

In the following chart we have 3 different `stackIds` this way we group 2 data keys by stackId `a`, 2 data keys by stackId `b` and the remaining data key is shown as a plain bar with the stackId `c`.

Screen Shot 2019-03-18 at 13 53 47


I don't know if we wish to support this, if we do, then it makes sense that for stackedbar charts the backend returns a `class / stackId` key per each `yn`.

sorodrigo:
> What I don't entirely understand is why meta->x, yn has a different format than meta->xAxis, yAxis, i.e. in the first case it contains the tooltip object which packages the suffix, prefix and format properties (leaving label outside), whereas in the latter all 4 properties are in the same level of the json. Not much difference really, but I'm not sure there's a good reason for it, so just wondering if it would be worth simplifying and formatting them in the same way.

Regarding this, we're currently using a fallback to the axes labels, because the tooltip ones are empty. I think maybe I was a bit ambitious separating both of them (would make sense if we wanted different labels in the tooltip than in the axes) maybe it's enough to have the axes ones and get rid of the `x, yn` objects? The only reason to have the `yn` would be if we want to specify `class/stackId` or similar data specific attributes as per my previous comment.

agnessa:
OK, I'll move type to axis meta 👍 

I think I might still be misunderstanding the tooltips, so maybe let's look at an example. Imagine we have a stacked bar chart:
- year on axis x,
- volume on axis y
- and break down by value of forest 500.

When we hover over a stack that refers to f500 value "5", what would the tooltip read - would it give the volume value for that year for that stack rather than f500 value? In such case, what should the "yn" metas look like; I thought they would be reflecting the possible values of f500, and so the unit would also be from f500:
```
    "y0": {"label": "1.0", "tooltip": {"prefix": "", "format": "", "suffix": "/5" }},
    "y1": {"label": "2.0", "tooltip": {"prefix": "", "format": "", "suffix": "/5" }},
    "y2": {"label": "3.0", "tooltip": {"prefix": "", "format": "", "suffix": "/5" }},
    "y3": {"label": "4.0", "tooltip": {"prefix": "", "format": "", "suffix": "/5" }},
    "y4": {"label": "5.0", "tooltip": {"prefix": "", "format": "", "suffix": "/5" }}
````

but in such case the suffix is incorrect if you want to show the volume, because it's different than the unit of the y axis. Maybe I'm mixing 2 things now:
1. a way to show tooltips
2. a way to render a legend

Let me know how you imagined the metas in this case and if anything else is needed for the legend.

BTW I noticed that sometimes the old chart endpoint returned 0-based "y" values and sometimes 1-based; I went for 0-based in the new endpoint (y0, y1, ...), but let me know if you prefer all of them to be 1-based.

agnessa:
I think the approach we've taken so far was for the frontend to handle response statuses rather than messages. I agree it would be good to improve the messages API-wide, especially when we start building the public-facing API; I think it's a bigger job.

agnessa:
I'll merge this, because there's already another PR that builds upon this, and it would be great to get feedback from FE while this is still fresh.

PR #737, opened 2019-03-18
Feat: dumbify filters – part I

sorodrigo:
This PR refactors current dropdown implementation to make it more flexible, it adds the clip last item feature and makes sure dropdown doesn't overflow the window.

~TODO~ For next PR:
- ~Use new dropdown for year selector.~
- ~Use new dropdown for recolor by selector.~
- ~Use new dropdown for nav generic selector.~

This PR is growing and I still need to add the dropdown variant for nav, dashboards, and use the new dropdown for the years, recolorby and resize by. I prefer splitting the PR into 2.

⚠️ This PR bumps React and Downshift to the latest versions. Hooks are here!


frontend/scripts/react-components/shared/dropdown/dropdown.component.jsx
```diff
@@ -6,7 +6,32 @@ import cx from 'classnames';
...
-  function renderItem(item, index, getItemProps) {
+  const listItemRef = useRef(null);
+  const [listHeight, updateListHeigh] = useState(null);
+  useEffect(() => {
+    if (listItemRef.current && listHeight === null && props.options && props.options.length > 0) {
+      const { height } = listItemRef.current.getBoundingClientRect();
+      const optionsLength = props.showSelected ? props.options.length : props.options.length - 1;
+      if (height > 0 && height * optionsLength > Dropdown.DEFAULT_MAX_LIST_HEIGHT) {
+        const newListHeight = height * 6 - height / 2;
+        updateListHeigh(newListHeight);
```
Bluesmile82: Typo here

frontend/scripts/react-components/shared/dropdown/dropdown.component.jsx
```diff
@@ -6,7 +6,32 @@ import cx from 'classnames';
...
+      }
+    }
+  });
```
Bluesmile82: Reading about hooks maybe we can limit the useEffect calls adding a second [listItemRef] attribute. We only want to trigger it the first time,  if it changes from null to the reference, don't we?: https://reactjs.org/docs/hooks-effect.html#tip-optimizing-performance-by-skipping-effects

Bluesmile82: Ok, reading a bit more we would have to include all the variables used in the scope so maybe is not helpful

frontend/scripts/react-components/shared/dropdown/dropdown.component.jsx
```diff
@@ -30,104 +60,161 @@ class Dropdown extends React.Component {
...
-              ) : null}
-            </Manager>
-          </div>
-        )}
-      </Downshift>
+      <ul
+        {...getMenuProps({
+          ref,
+          style: styleToApply,
+          'data-placement': placement,
```
Bluesmile82: image
I think moving some extra pixels to the top would be needed

Bluesmile82:
Maybe we can use 'exhaustive-deps' lint rule. Seems interesting for UseEffect
 https://github.com/facebook/react/issues/14920

frontend/scripts/react-components/shared/dropdown/dropdown.component.jsx
sorodrigo: A way of only executing once is to pass an empty array, this would be equivalent to `componentDidMount` however I think we need to run it at least twice so we’re screwed there.

You should only add as dependencies the values that cause a re-render so only state vars and props. Adding a ref is not recommended.

Because of the if statement this should only call the function once.

frontend/scripts/react-components/shared/dropdown/dropdown.component.jsx
```diff
@@ -30,78 +30,94 @@ class Dropdown extends React.Component {
...
+    variant,
+    readOnly,
+    placement,
+    options,
+    children
+  } = props;
+
+  // popper won't detect changes on its children so the necessary recalculations won't happen
+  // we create a key that changes every time the options or children change, that way we remount the popper component
+  const popperForceUpdateKey = useRef(0);
```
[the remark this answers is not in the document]
sorodrigo: Let’s have it in mind for our next popper implementation. Maybe then it will be easier to see how.

Bluesmile82:
Thanks for the changes!

PR #738, opened 2019-03-18
Add role attribute to ContextNodeTypeProperty

aabdaab:
* Add role attr to context_node_type_properties
* Add validations on DB and AR levels
* Apply changes to CMS


agnessa:
Apologies for the mistake in task description - the 4th role name should be "destination". I amended the migration and added 2 sql updates which set the default values: first one sets it according to the `column_group` attribute, and the second one nullifies it for some of the node types. I use this king of select to review the assignment:

```
SELECT context_id, node_type_id, node_types.name, column_position, column_group, role
FROM context_node_type_properties
JOIN context_node_types ON context_node_type_properties.context_node_type_id = context_node_types.id
JOIN node_types ON context_node_types.node_type_id = node_types.id
ORDER BY context_id, column_position
```

Please check if that makes sense for you, and run `db:migrate:redo` locally.

Otherwise it looks good to me.

PR #739, opened 2019-03-19
Feature/horizontal chart

Bluesmile82:
- Add horizontal bar and horizontal stacked bar chart to docs
- Fix some warnings and refactor a bit the charts component

Note: Some details are missing like fixing the problem with the long texts in the x-axis. This could be solved with an ellipsis and a tooltip (react-popper) and maybe some legend for mobile versions


frontend/scripts/react-components/dashboard-element/dashboard-widget/dashboard-widget-config.js
```diff
@@ -78,6 +78,32 @@ export default {
...
+    xKeys: {
+      bars: null
+    },
+    yKey: 'y',
+    xKeysAttributes: {
+      bars: {
+        interval: 'preserveStartEnd'
+      }
+    },
+    yAxis: {
```
sorodrigo: This comes from the API in meta

frontend/scripts/react-components/chart/chart.component.jsx
```diff
@@ -128,28 +115,8 @@ class Chart extends PureComponent {
...
-                    unit={unit || ''}
-                    unitFormat={unitFormat || (value => value)}
-                    fill="#555555"
-                  />
-                }
-                {...yAxis}
-              />
-            )}
-
+            {CustomXAxis({ config, data })}
```
sorodrigo: Why not mount it as a component?

Bluesmile82: I tried but the ComposedChart in recharts only accept some specific components. If you include something else it doesn't render it without a warning

Bluesmile82:
Ready again! Let's fix the cursor in the chart style fixes PR

PR #745, opened 2019-03-22
Feat/dropdown refacto

sorodrigo:
This PR updates the dropdown to include the functionality that the tool dropdowns have. It also replaces the dropdowns in the sankey and logistics map. I haven't done any profiling but this should come with some performance improvements and of course better accesibility.


sorodrigo:
Fixed 🏁

PR #747, opened 2019-03-22
Apply node type role attribute to dashboards filters

aabdaab:
This PR applies a new **role** attribute from _context_node_type_properties_ to dependent materialized views.
I also updated the specs although two of them are still failing; one has something to do with populating materialized views and the other returns one more company than before - @agnessa could you steer me in the right direction what may be going on in here?

### DIFF
--- a/app/models/api/v3/readonly/dashboards/country.rb
+++ b/app/models/api/v3/readonly/dashboards/country.rb
@@ -9,15 +9,6 @@
 #  commodity_id  :integer
 #  node_id       :integer
 #
-# Indexes
-#
-#  dashboards_countries_mv_commodity_id_idx   (commodity_id)
-#  dashboards_countries_mv_group_columns_idx  (id,name)
-#  dashboards_countries_mv_name_idx           (name)
-#  dashboards_countries_mv_name_tsvector_idx  (name_tsvector) USING gin
-#  dashboards_countries_mv_node_id_idx        (node_id)
-#  dashboards_countries_mv_unique_idx         (id,node_id,commodity_id) UNIQUE
-#
 
 module Api
   module V3
--- a/app/models/api/v3/readonly/dashboards/country.rb
+++ b/app/models/api/v3/readonly/dashboards/country.rb
@@ -9,6 +9,15 @@
 #  commodity_id  :integer
 #  node_id       :integer
 #
+# Indexes
+#
+#  dashboards_countries_mv_commodity_id_idx   (commodity_id)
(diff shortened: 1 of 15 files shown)


agnessa:
The spec which returns one more company than before - in fact it's a port rather than company, so I'd suggest setting the role to `nil` for that node type. As for the materialized views errors, you need to refresh the materialized view in the `before` block, but if in the same block it's already refreshing `Api::V3::Readonly::Attribute` then be careful to refresh them in a way that doesn't introduce a circular refresh, e.g. this is a similar situation:

```
Api::V3::Readonly::Attribute.refresh(sync: true, skip_dependents: true)
Api::V3::Readonly::ChartAttribute.refresh(sync: true, skip_dependencies: true)
```

aabdaab:
^ Fixed
I also changed the base to `develop`

db/structure.sql
```diff
@@ -3006,7 +3005,7 @@ CREATE MATERIALIZED VIEW public.dashboards_flow_paths_mv AS
      JOIN public.node_types ON ((nodes.node_type_id = node_types.id)))
      JOIN public.context_node_types cnt ON (((node_types.id = cnt.node_type_id) AND (flow_paths.context_id = cnt.context_id))))
      JOIN public.context_node_type_properties cnt_props ON ((cnt.id = cnt_props.context_node_type_id)))
-  WHERE ((cnt_props.column_group = ANY (ARRAY[0, 3])) OR (node_types.name = ANY (ARRAY['COUNTRY'::text, 'IMPORTER'::text, 'EXPORTER'::text, 'TRADER'::text])))
+  WHERE (((cnt_props.role)::text = ANY ((ARRAY['source'::character varying, 'exporter'::character varying, 'importer'::character varying, 'destination'::character varying])::text[])) OR (node_types.name = ANY (ARRAY['COUNTRY'::text, 'IMPORTER'::text, 'EXPORTER'::text, 'TRADER'::text])))
```
agnessa: I think maybe we don't need this condition anymore; perhaps it's equivalent to check if role is not null?

app/models/api/v3/readonly/dashboards/country.rb
```diff
@@ -9,15 +9,6 @@
 #  commodity_id  :integer
 #  node_id       :integer
 #
-# Indexes
```
agnessa: Any idea why this disappeared? I think in the migration you're recreating the indexes correctly.

aabdaab:
Thanks for noticing 🙌

PR #754, opened 2019-03-27
Fix/swap x and y for horizontal bar charts

agnessa:


sorodrigo:
Screen Shot 2019-03-27 at 13 06 58

sorodrigo:
Thanks

PR #757, opened 2019-03-27
Multi year, ncont, node type view (stacked bar chart)

agnessa:
This doesn't add a new endpoint - reuses the same endpoint as for the "overview". The difference is in how the endpoint is filtered, the idea being these charts only make sense when flow any path filters are selected.


agnessa:
Let me fix the style issues

PR #760, opened 2019-03-28
Fix/one chart position

Bluesmile82:
This PR adds a No selection option to the recolour by and fixes the width problems with one chart

image
image


frontend/scripts/react-components/dashboard-element/dashboard-element.scss
```diff
@@ -27,7 +27,7 @@
   }
 
   .dashboard-element-widgets {
-    padding: 0 0 40px;
+    padding: 0 20px 40px;
```
sorodrigo: have you tested this with a title? probably explodes

frontend/scripts/react-components/dashboard-element/dashboard-element.component.jsx
```diff
@@ -93,7 +93,7 @@ class DashboardElement extends React.PureComponent {
     return (
       <div className="row -equal-height -flex-end">
         {charts.map(chart => (
-          <div key={chart.id} className="column small-12 medium-6 ">
+          <div key={chart.id} className="column medium-12 large-6 ">
```
sorodrigo: I'm not sure about this, in tablets all charts will show in one line

PR #763, opened 2019-04-01
Set bars min height

Bluesmile82:
- Set bars min height to solve the problem of crowded horizontal bar charts
- Remove cartesian grid for donut chart

image


sorodrigo:
could we add it to the config instead of the component?

PR #765, opened 2019-04-01
Add special splash screen on IE

sorodrigo:
Adds a splash screen that will show when using IE 11 or lower.


sorodrigo:
Screen Shot 2019-04-02 at 18 26 36

frontend/scripts/index.ie.ejs
```diff
@@ -33,19 +33,19 @@
...
       <div class="column small-12 medium-3">
-        <a class="browser-box">
+        <a class="browser-box" href="https://www.mozilla.org/en-US/firefox/new/">
           <span class="browser-title">Firefox</span>
           <img class="browser-logo" src="/images/browsers/firefox.svg" />
         </a>
       </div>
       <div class="column small-12 medium-3">
-        <a class="browser-box">
+        <a class="browser-box" href="https://www.microsoft.com/en-us/windows/microsoft-edge">
```
Bluesmile82: Maybe we could use the language here, or will it be too much?

sorodrigo: I think this is good enough for now

PR #767, opened 2019-04-02
Fix for mixed up biomes in sankey filter dropdown

agnessa:
https://www.pivotaltracker.com/story/show/165055602

### DIFF
--- a/app/models/api/v3/context.rb
+++ b/app/models/api/v3/context.rb
@@ -71,6 +71,30 @@ module Api
         end
       end
 
+      def biome_nodes
+        biome_context_node_type = context_node_types.
+          joins(:node_type).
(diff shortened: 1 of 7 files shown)


app/models/api/v3/context.rb
```diff
@@ -71,6 +71,30 @@ def exporter_context_node_type
         end
       end
 
+      def biome_nodes
+        biome_context_node_type = context_node_types.
+          joins(:node_type).
+          where('node_types.name' => NodeTypeName::BIOME).
```
aabdaab: if there's only one we could use `find_by`

agnessa: 👍

PR #771, opened 2019-04-03
Fix: makes meta data keys validation stronger

sorodrigo:
This PR checks that the data keys are actually data keys.


sorodrigo:
I feel so smart using regex 🤓 so please try to break it 🙏

PR #773, opened 2019-04-03
Add ranking chart logic

aabdaab:
PT task: https://www.pivotaltracker.com/story/show/165025068

**Issue**: 
Currently on the dashboard (`/dashboards/new?`) some data is not suitable for `HORIZONTAL_BAR_CHART` or `HORIZONTAL_STACKED_BAR_CHART` because `Other` is way bigger than the rest of the values. I added a threshold (20%) allowed to return these default chart types, otherwise I'm returning `ranking-chart`.
The change is visible in `parametrised_charts` endpoint.


sorodrigo:
Hey! Great job! We have a situation, we can't show the ranking when a non-continuous indicator is selected because there's no way of showing the different classes.

Also after discussing with Ben he told me that we should show the value compared to the total. Could you send us the total value as another dimension? Thanks!

aabdaab:
no worries!
just to be clear, for the total value do you want all the entities summed including `OTHER` if other's present? (Below you've got an example of data with one `y0`, so the total value would mean for all the `y0`).
image
Also, if that's the case, when it is stacked bar chart, you want a total value per each `y`; y0, y1, y2..., is that correct?
I would suggest the API response would be as follows:
```
# endpoint /parametrised_charts
data: [ objects with types and urls ],
total_value: { y0: 12345, y1, y2 if this is stacked bar chart  }
```

aabdaab:
also, another thing I just realized that I took for granted... for the parametrised_charts I do not get a `top_n` query param from FE. I assumed it will be `10` but not sure if it's correct to take that assumption. I would suggest passing that as a query param as in the charts endpoints

aabdaab:
☝️added the `aggregates` key to `meta` object where for now there's only `total_value`. This change affects 3 endpoints: `multi_year_no_ncont_node_type_view`, `single_year_ncont_node_type_view`, `single_year_no_ncont_node_type_view`
Also this method in `parametrized_charts` 
```
def single_year_ncont_node_type_view(node_type)
          {
            source: :single_year_ncont_node_type_view,
            type: HORIZONTAL_STACKED_BAR_CHART,
            x: :node_type,
            break_by: :ncont_attribute,
            node_type_id: node_type.id
          }
        end
```
went back to what it looked like before; it doesn't take into account ranking-chart option. Let me know if that's correct, then I will remove the helper services that were checking against that possibility 👌

sorodrigo:
It's great thank you!

agnessa:
Hum, I'm not a huge fan of this solution, as it slows down the `parametrised_charts` endpoint for some parameters, because of the necessity to run the chart queries in order to get the totals. I thought this could be avoided by letting the frontend decide which chart to show, as I mentioned to Alvaro. E.g. you get a bar chart, check OTHER is too big for a bar chart and show the ranking chart, omitting OTHER. We'd need to add pagination and totals to the chart endpoint to make it fully compatible, and we could also change parametrised_charts to return 2 possible chart types to display instead of one, but that would be a much smaller change. I imagine this is now too late to shift direction, but I think going forward we need to have backend performance at heart. The flows queries are quite slow and will get slower with more data added, which is why they were separated into their own distinct endpoints to allow loading them in parallel by frontend and improving caching efficiency; and some of them are now run twice.

sorodrigo:
Sorry I wasn’t aware of this. For the frontend the solution you suggested of receiving 2 chart types would work. But like you said, if we don’t have pagination we need to refetch the data again (and the way is built this will complicate things a bit).

Does the way of calculating the aggregations (total) also needs to run queries twice? What’s your suggestion there?

agnessa:
No, the aggregations don't add a new query, that part is fine.

sorodrigo:
Ok then we should move the logic that determines what to paint to the frontend like you suggest. If the endpoints are the same for both charts, we don’t really need to receive 2 chart types. And for now, we can fetch twice only on the charts that need it. It’s better than the current solution and could be improved with pagination if needed.

sorodrigo:
After discussing with @agnessa we decided to leave it as is for now. And optimize later, because we might end up removing the ranking or it could end up changing.

PR #774, opened 2019-04-04
Feature/table data widget

Bluesmile82:
This is the second PR for the table widget:

image
The different cases could be found here: https://docs.google.com/document/d/13NH3nkeMNWsdbwo2bzHWH8hhvf7AlgliMz8NNM2kYIU/

Note: The last three charts are not coming from the backend

Extra:

- Fix legend style


frontend/scripts/react-components/dashboard-element/dashboard-widget/table-modal/table/table.component.jsx
```diff
@@ -0,0 +1,75 @@
...
+        rowCount={data.length}
+        columnCount={headers.length}
+      >
+        {({ rowIndex, columnIndex, style }) => {
+          const item = data[rowIndex][columnIndex];
+          if (typeof item === 'undefined') return null;
+          return (
+            <td
+              style={style}
+              className={cx('header-item', rowIndex % 2 ? 'list-item-odd' : 'list-item-even')}
```
sorodrigo: there a css pseudo-selector for this. https://developer.mozilla.org/en-US/docs/Web/CSS/:nth-child#Functional_notation

frontend/scripts/react-components/dashboard-element/dashboard-widget/table-modal/table/table.component.jsx
```diff
@@ -0,0 +1,75 @@
...
+            >
+              <Text size="md" as="span" variant="mono">
+                {item}
+              </Text>
+            </td>
+          );
+        }}
+      </VariableSizeGrid>
+      {loading && (
```
sorodrigo: I wonder if we could use a placeholder instead? We know the height and width. Check out the `<Top />` component.

frontend/scripts/react-components/table-modal/table-modal.selectors.js
```diff
@@ -0,0 +1,98 @@
...
+  const unit = type === 'cont_attribute' ? meta.xAxis.suffix : meta.yAxis.suffix;
+  return `${meta.info.filter[type].toUpperCase()}${unit ? ` (${unit})` : ''}`;
+};
+
+const getVariableColumnName = (d, c, meta) => d.y || meta[c].label;
+
+const getCommodities = info => info.commodities[0];
+const getCountries = info => info.countries[0];
+
+export const getTableData = createSelector(
```
sorodrigo: should be a factory

frontend/scripts/react-components/dashboard-element/dashboard-widget/dashboard-widget.component.jsx
```diff
@@ -12,14 +15,38 @@ import RankingWidget from 'react-components/ranking-widget';
...
-  const { title, loading, error, data, chartConfig, dynamicSentenceParts } = props;
+  const { title, loading, error, data, meta, chartType, chartConfig, dynamicSentenceParts } = props;
 
   const renderError = errorMessage => (
     <Text color="white" weight="bold" variant="mono" size="lg" className="widget-centered">
       {errorMessage}
     </Text>
   );
 
+  const [isModalOpen, openModal] = useState(false);
```
sorodrigo: Yay 🎉 ! First hook!

frontend/scripts/react-components/dashboard-element/dashboard-widget/dashboard-widget.component.jsx
```diff
@@ -12,14 +15,38 @@ import RankingWidget from 'react-components/ranking-widget';
...
+        {hasTable && (
+          <>
+            <Button
+              icon="icon-table"
+              color="charcoal"
+              variant="circle"
+              onClick={() => openModal(true)}
+              disabled={loading}
+            />
+            <SimpleModal isOpen={isModalOpen} onClickClose={() => openModal(false)}>
```
sorodrigo: I think we should use `onRequestClose` instead

frontend/scripts/react-components/dashboard-element/dashboard-widget/dashboard-widget.container.jsx
```diff
@@ -26,14 +31,31 @@ class DashboardWidgetContainer extends Component {
...
       <DashboardWidgetComponent
         data={data}
+        id={id} // This may be substituted by the title if its unique
+        meta={meta}
         title={title}
         error={error}
         loading={loading || chartsLoading}
         chartConfig={this.addTooltipContentToConfig(config, meta)}
+        chartType={chartType}
+        setActiveModal={setActiveModal}
```
sorodrigo: not needed anymore right?

frontend/scripts/react-components/dashboard-element/dashboard-widget/table-modal/table-modal.scss
```diff
@@ -0,0 +1,18 @@
+@import 'styles/settings';
+
+.c-table-modal {
+  width: 820px;
+  padding: 30px 0;
+  display: flex;
+  flex-direction: column;
```
sorodrigo: Do we need this? Most of the times `display: block;` and `margin: 0 auto` would do the trick. I'm saying this, because recently I realised I abuse `flex-direction: column`.

https://css-tricks.com/centering-css-complete-guide/

frontend/scripts/react-components/dashboard-element/dashboard-widget/table-modal/table-modal.selectors.js
```diff
@@ -0,0 +1,149 @@
...
+const hasNContIndicator = createSelector(
+  [getMeta],
+  meta => meta.info.filter.ncont_attribute
+);
+const hasMultipleYears = createSelector(
+  [getMeta],
+  meta => meta.info.years.end_year
+);
+const hasVariableColumn = createSelector(
+  [getChartType, getVariableColumnName, hasMultipleYears, hasNContIndicator, getSourcesPanel],
```
sorodrigo: I usually call the fns as `getHasMultipleYears` that way the value doesn't need the underscore. That said, there's no need to change this.

frontend/scripts/react-components/dashboard-element/dashboard-widget/table-modal/table-modal.selectors.js
```diff
@@ -0,0 +1,149 @@
...
+  returnNameArrayIfNotEmpty(
+    state.dashboardElement && state.dashboardElement.countriesPanel.activeItems
+  );
+const getCommoditiesPanel = state =>
+  returnNameArrayIfNotEmpty(
+    state.dashboardElement && state.dashboardElement.commoditiesPanel.activeItems
+  );
+const getSourcesPanel = state =>
+  returnNameArrayIfNotEmpty(
+    state.dashboardElement && state.dashboardElement.commoditiesPanel.activeItems
```
sorodrigo: _Did you mean "sourcesPanel" ?_ 😛

frontend/scripts/react-components/dashboard-element/dashboard-widget/table-modal/table-modal.selectors.js
```diff
@@ -0,0 +1,149 @@
...
+  meta => meta.info.filter.ncont_attribute
+);
+const hasMultipleYears = createSelector(
+  [getMeta],
+  meta => meta.info.years.end_year
+);
+const hasVariableColumn = createSelector(
+  [getChartType, getVariableColumnName, hasMultipleYears, hasNContIndicator, getSourcesPanel],
+  (chartType, variableColumnName, _hasMultipleYears, _hasNContIndicator, sourcesPanel) =>
+    !(!sourcesPanel && _hasNContIndicator && _hasMultipleYears) &&
```
sorodrigo: For me is strange that you have to check the sourcesPanel value, because it's a mandatory value.

frontend/scripts/react-components/dashboard-element/dashboard-widget/table-modal/table-modal.selectors.js
```diff
@@ -0,0 +1,149 @@
...
+import isEmpty from 'lodash/isEmpty';
+import { format } from 'd3-format';
+import { CHART_TYPES } from 'constants';
+
+const getMeta = (state, { meta }) => meta || null;
+const getData = (state, { data }) => data || null;
+const getChartType = (state, { chartType }) => chartType || null;
+const returnNameArrayIfNotEmpty = value =>
+  isEmpty(value) ? null : Object.values(value).map(v => v.name);
+const getCountriesPanel = state =>
```
sorodrigo: The name is a bit off IMO, you expect a panel and instead you get an activeItem.

frontend/scripts/react-components/dashboard-element/dashboard-widget/table-modal/table-modal.selectors.js
```diff
@@ -0,0 +1,149 @@
...
+  (chartType, variableColumnName, _hasMultipleYears, _hasNContIndicator, sourcesPanel) =>
+    !(!sourcesPanel && _hasNContIndicator && _hasMultipleYears) &&
+    !(!sourcesPanel && chartType === 'bar' && _hasMultipleYears) &&
+    !(chartType === 'pie') &&
+    variableColumnName
+);
+
+const getIndicatorColumnName = (meta, type) => {
+  const unit = type === 'cont' && (meta.xAxis.suffix || meta.yAxis.suffix);
+  return `${meta.info.filter[`${type}_attribute`].toUpperCase()}${unit ? ` (${unit})` : ''}`;
```
sorodrigo: I generally don't like using uppercase strings, prefer having the CSS style it. Mainly because if you need to update to capitalised you have to track this transform and change it back.

frontend/scripts/react-components/dashboard-element/dashboard-widget/table-modal/table-modal.selectors.js
```diff
@@ -0,0 +1,149 @@
...
+);
+
+const getYear = (d, meta, chartType) => {
+  switch (chartType) {
+    case CHART_TYPES.horizontalBar:
+    case CHART_TYPES.horizontalStackedBar:
+    case CHART_TYPES.pie:
+      return meta.info.years.start_year;
+    default:
+      return d.x || meta.info.years.start_year;
```
sorodrigo: I suppose this will change once we fix overview charts?

frontend/scripts/react-components/dashboard-element/dashboard-widget/table-modal/table-modal.selectors.js
```diff
@@ -0,0 +1,149 @@
...
+  }
+};
+
+const getContValue = (d, c) => d[c] && format(',.2s')(d[c]);
+const getNotContValue = (d, c, meta, chartType, _hasMultipleYears) => {
```
sorodrigo: Should be `nonContValue`

frontend/scripts/react-components/dashboard-element/dashboard-widget/table-modal/table-modal.selectors.js
```diff
@@ -0,0 +1,149 @@
...
+    const metaColumns = {
+      ...meta
+    };
+    const notColumnKeys = ['info', 'xAxis', 'yAxis'];
+    if (chartType === CHART_TYPES.horizontalBar || chartType === CHART_TYPES.horizontalStackedBar) {
+      notColumnKeys.push('y');
+    } else {
+      notColumnKeys.push('x');
+    }
+    notColumnKeys.forEach(a => delete metaColumns[a]);
```
sorodrigo: could we use `omit`?

frontend/scripts/react-components/dashboard-element/dashboard-widget/table-modal/table-modal.selectors.js
```diff
@@ -0,0 +1,149 @@
...
+const getMeta = (state, { meta }) => meta || null;
+const getData = (state, { data }) => data || null;
+const getChartType = (state, { chartType }) => chartType || null;
+const returnNameArrayIfNotEmpty = value =>
+  isEmpty(value) ? null : Object.values(value).map(v => v.name);
+const getCountriesPanel = state =>
+  returnNameArrayIfNotEmpty(
+    state.dashboardElement && state.dashboardElement.countriesPanel.activeItems
+  );
+const getCommoditiesPanel = state =>
```
sorodrigo: Should this and the countries be an array?

frontend/scripts/react-components/dashboard-element/dashboard-widget/table-modal/table-modal.selectors.js
```diff
@@ -0,0 +1,149 @@
...
+};
+
+const getContValue = (d, c) => d[c] && format(',.2s')(d[c]);
+const getNotContValue = (d, c, meta, chartType, _hasMultipleYears) => {
+  if (
+    chartType === CHART_TYPES.horizontalBar ||
+    chartType === CHART_TYPES.horizontalStackedBar ||
+    _hasMultipleYears
+  ) {
+    return meta[c].label && meta[c].label.toUpperCase();
```
sorodrigo: what does `c` stands for? I prefer using less abbreviations when possible

frontend/scripts/react-components/dashboard-element/dashboard-widget/table-modal/table-modal.selectors.js
```diff
@@ -0,0 +1,149 @@
...
+    _hasVariableColumn,
+    _hasMultipleYears
+  ) => {
+    if (!meta || !data) return null;
+    const parsedData = [];
+    dataColumnNames.forEach(c => {
+      data.forEach(d => {
+        const rowData = [countries[0], commodities[0], getYear(d, meta, chartType)];
+        if (_hasVariableColumn) {
+          rowData.push(getVariableColumnData(d, c, meta));
```
sorodrigo: same d & c

frontend/scripts/react-components/dashboard-element/dashboard-widget/table-modal/table/table.scss
```diff
@@ -0,0 +1,25 @@
+@import 'styles/settings';
+
+.c-table {
+  .table-header {
+    display: flex;
+    flex-direction: row;
+    justify-content: space-evenly;
```
sorodrigo: I think `space-evenly` is not supported on `flex` in Edge. Also `flex-direction: row;` is not needed. Maybe we need to use `display: grid;`? We can, because we don't support IE 😬

frontend/scripts/react-components/table-modal/table-modal.component.jsx
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Heading from 'react-components/shared/heading/heading.component';
+import Button from 'react-components/shared/button';
+import Table from 'react-components/dashboard-element/dashboard-widget/table-modal/table';
+import 'react-components/dashboard-element/dashboard-widget/table-modal/table-modal.scss';
```
sorodrigo: This file is not needed right?

frontend/scripts/react-components/dashboard-element/dashboard-widget/table-modal/table-modal.selectors.js
```diff
@@ -0,0 +1,149 @@
...
+      notColumnKeys.push('y');
+    } else {
+      notColumnKeys.push('x');
+    }
+    notColumnKeys.forEach(a => delete metaColumns[a]);
+    return Object.keys(metaColumns);
+  }
+);
+
+export const getTableData = createSelector(
```
sorodrigo: I think this should be a factory

frontend/scripts/react-components/dashboard-element/dashboard-widget/table-modal/table-modal.scss
Bluesmile82: I think flex is a better way of positioning. Also in fluid structures. Not sure how this is going to change

frontend/scripts/react-components/dashboard-element/dashboard-widget/table-modal/table-modal.selectors.js
Bluesmile82: See the next comment, the names have to improve

sorodrigo: But still, if the name was `sourcesPanelActiveItem` or other. This check is strange. In order to have a widget, we need `sourcesPanelActiveItem` to exist

frontend/scripts/react-components/dashboard-element/dashboard-widget/table-modal/table-modal.selectors.js
Bluesmile82: I think is even more obscured in the CSS 😟

frontend/scripts/react-components/dashboard-element/dashboard-widget/table-modal/table-modal.selectors.js
sorodrigo: Nevermind, I realised that sourcesPanel is not countriesPanel. My bad :v:

frontend/scripts/react-components/dashboard-element/dashboard-widget/table-modal/table-modal.selectors.js
sorodrigo: IMO CSS should take care of presentation, values shouldn't come in a presentational way.

frontend/scripts/react-components/dashboard-element/dashboard-widget/table-modal/table-modal.selectors.js
Bluesmile82: They are an array 🤷‍♂️

frontend/scripts/react-components/dashboard-element/dashboard-widget/table-modal/table-modal.selectors.js
Bluesmile82: d is a unit of data, can you find a better name?

frontend/scripts/react-components/table-modal/table-modal.component.jsx
Bluesmile82: The SCSS? We need it

frontend/scripts/react-components/dashboard-element/dashboard-widget/table-modal/table-modal.scss
sorodrigo: Why is it better in this case? I think `flex-direction: column` should be only used if needed: it adds more to it's children than just centering.

frontend/scripts/react-components/dashboard-element/dashboard-widget/table-modal/table/table.scss
Bluesmile82: Actually, we don't need this 😆

frontend/scripts/react-components/table-modal/table-modal.component.jsx
sorodrigo: Nope sorry, the whole `react-components/table-modal` directory.

frontend/scripts/react-components/dashboard-element/dashboard-widget/table-modal/table-modal.selectors.js
sorodrigo: exactly haha. Sorry I meant to say: Shouldn't this be an item or null.

frontend/scripts/react-components/table-modal/table-modal.component.jsx
Bluesmile82: Is true 🤦‍♂️

frontend/scripts/react-components/dashboard-element/dashboard-widget/table-modal/table/table.component.jsx
Bluesmile82: We can't use it as we need the row index. React window displays directly the cells

frontend/scripts/react-components/dashboard-element/dashboard-widget/table-modal/table-modal.selectors.js
Bluesmile82: ActiveItems is an array in all cases. As the purpose is to have several countries and commodities I think it makes sense to keep them all as arrays. Nevertheless we will always have a country and a commodity

frontend/scripts/react-components/dashboard-element/dashboard-widget/table-modal/table/table.component.jsx
sorodrigo: I see. I forgot about this 🤦‍♂️ . If we're putting all rows inside a single `<tr />` we lose all the table advantages. Doing some research I found that I was also wrong about the table element thing. https://github.com/bvaughn/react-window/issues/60

Apologies, I guess we shouldn't use a table then.

Bluesmile82: Yeah, I also thought that about the tr. I will change it to divs and we will get rid of the warnings

frontend/scripts/react-components/dashboard-element/dashboard-widget/table-modal/table-modal.selectors.js
sorodrigo: Not sure I love being this smart into the future, but I can understand your point. Oh well, let's leave it as is.

frontend/scripts/react-components/dashboard-element/dashboard-widget/table-modal/table-modal.selectors.js
Bluesmile82: In this specific case is quite difficult as we will have to add a span and that will be ugly as this is a selector. Not all columns will have unit and that will add much more code to the component. Do you mind if we keep it like this or you have some alternative?

sorodrigo: You could return an object instead of a string. I do think that this is important.

`{ name: ‘trade volume’, unit: ‘t’ }`

Bluesmile82:
Ready again!

Bluesmile82:
image

Now with Sorting ™️

PR #781, opened 2019-04-11
Feat/intersection observer

sorodrigo:
This PR uses the `IntersectionObserver` API to avoid rendering stuff below the fold. It brings amazing speed improvements!! 😬 

This is a capture showing the lighthouse analysis for home page with IntersectionObserver (top) and without it (bottom). Note that the Best Practices score drops a bit because I'm serving the app from a unsecure connection (my localhost).

**First Meaningful Paint drops by 3secs!!!!**

> "For every 100 ms of improved website speed, Walmart was able to grow incremental revenue by up to 1%."

Screen Shot 2019-04-11 at 11 32 21

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -148,6 +148,7 @@
     "react-dom": "^16.8.6",
     "react-iframe-resizer-super": "^0.2.0",
     "react-input-autosize": "^2.2.1",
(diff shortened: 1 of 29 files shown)


tiagojsag:
I would suggest a feature flag

PR #783, opened 2019-04-11
Ports were not showing in the dynamic sentence

Bluesmile82:
image

### DIFF
--- a/frontend/scripts/react-components/dashboard-element/dashboard-element.selectors.js
+++ b/frontend/scripts/react-components/dashboard-element/dashboard-element.selectors.js
@@ -105,6 +105,15 @@ export const getDynamicSentence = createSelector(
       },
       {
         panel: 'companies',
+        nodeTypeName: 'ports',
+        id: 'port',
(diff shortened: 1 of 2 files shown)


frontend/scripts/react-components/dashboard-element/dashboard-element.selectors.js
```diff
@@ -103,6 +103,15 @@ export const getDynamicSentence = createSelector(
         value: sourcesValue,
         transform: 'capitalize'
       },
+      {
+        panel: 'companies',
+        nodeTypeName: 'ports',
+        id: 'port',
+        prefix: getActivePanelItem('companies', 'PORT') ? 'exported from' : '',
```
agnessa: Please bear in mind that the node types enabled for each of the panels are context-dependent and their names might change / new node types might be added. This won't happen in the nearest data release I believe, so maybe this is not for now, but it would be ideal if we didn't need to hardcode any node type names and instead base this logic on the node type names returned by the filter_meta endpoint.

sorodrigo:
I'll close this until we find a better solution.

PR #789, opened 2019-04-12
Fix/unit and si sorting

Bluesmile82:
This PR fixes two bugs:

- The unit was wrong for some indicators in the table (We are still missing this information from the backend in horizontal and horizontal stacked bar charts)

- The sorting was wrong when we had k and M SI suffix in the numbers of the table. Now they are translated back into full integers and sorted this way


frontend/scripts/tests/table-modal/table-modal-selectors.test.js
```diff
@@ -45,30 +45,36 @@ describe('Table widget selectors', () => {
...
           meta: mYearNCont,
           data,
           chartType
         })
       ).toEqual([
-        { name: 'COMMODITY' },
-        { name: 'COUNTRY' },
-        { name: 'YEAR' },
-        { name: 'TRADE VOLUME', unit: 't' },
-        { name: 'BIOMES' },
```
Bluesmile82: This is supposed to be only for the last case. That is only fixed with the last backend PR

PR #790, opened 2019-04-16
Feature/dashboard e2e tests

Bluesmile82:
This PR creates the dashboard e2e tests.

Also integrates a new workflow for creating the e2e tests with: [Polly](https://netflix.github.io/pollyjs/#/) and [Puppeteer recorder](https://github.com/checkly/puppeteer-recorder). The docs are updated with more information. But basically we won't need the mocks anymore as they will be recorded


sorodrigo:
Great work!

PR #792, opened 2019-04-23
Fix/legend ellipsis

Bluesmile82:
image

- Create Ellipsis component and add it to widget legend
- Fix docz
- Added tooltip component with react-popper
- Added media query to control page size in legend


Bluesmile82:
Updated!

Bluesmile82:
We can actually merge this one. The improved version will come later

sorodrigo:
It still overflows on mobile resolution, let's wait for the redesign.

PR #794, opened 2019-04-24
Fix/chart tooltip

Bluesmile82:
image

Design is still in process

I think we are not getting the unit from the backend

### DIFF
--- a/frontend/scripts/react-components/dashboard-element/dashboard-widget/dashboard-widget-tooltip.jsx
+++ b/frontend/scripts/react-components/dashboard-element/dashboard-widget/dashboard-widget-tooltip.jsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import Text from 'react-components/shared/text';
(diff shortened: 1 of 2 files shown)


sorodrigo:
It's still ordered by amount instead of class. 🙈

Bluesmile82:
Sorry, I forgot to push 😁

frontend/scripts/react-components/dashboard-element/dashboard-widget/dashboard-widget-tooltip.jsx
```diff
@@ -41,21 +42,45 @@ class DashboardWidgetTooltip extends React.PureComponent {
...
+            variant="mono"
+            as="span"
+            color="white"
+            weight="bold"
+            className="dashboard-widget-tooltip-unit"
+          >
+            {payload[0] && payload[0].unit}
+          </Text>
+        </div>
+        {payload.reverse().map(item => (
```
sorodrigo: payload mutates the array we'll need to make a copy beforehand. I'll do it

PR #796, opened 2019-04-25
Fix iframe resizing

Bluesmile82:
Fix the scrollbar appearing in some cases in the GFW iframe on a place profile

We use react-iframe-resizer-super. The only drawback is that callbacks won't work in different host. There is an issue opened: https://github.com/zeroasterisk/react-iframe-resizer-super/issues/14 

Extra: Fix icon in sankey column dropdown (Checked the other icons in the page and they are 👌 )
image


Bluesmile82:
Updated!

PR #801, opened 2019-04-30
Use a middleware to add the current Language to url when we switch pages

Bluesmile82:
Use a middleware to add the current Language to url when we switch pages. Its in the after change as in the before change we still not have the variable.

The default is 'en' and we can't use Transifex as its not loading in this hook

### DIFF
--- a/frontend/scripts/router/router.js
+++ b/frontend/scripts/router/router.js
@@ -165,7 +165,6 @@ const config = {
 
     return dispatchThunks(redirectToExplore)(dispatch, getState, { action });
   },
-  // eslint-disable-next-line consistent-return
   onAfterChange: (dispatch, getState, { action }) => {
     const currentLanguage = action.meta.location?.current?.query?.lang;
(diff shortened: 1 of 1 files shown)


frontend/scripts/router/router.js
```diff
@@ -165,6 +165,17 @@ const config = {
...
   },
+  // eslint-disable-next-line consistent-return
+  onAfterChange: (dispatch, getState, { action }) => {
+    const currentLanguage = action.meta.location?.current?.query?.lang;
+    const previousLanguage = action.meta.location?.prev?.query?.lang;
+    if (!currentLanguage) {
+      const { location } = getState();
+      const query = { ...location.query, lang: previousLanguage || 'en' };
+      const payload = { ...location.payload, query };
+      return dispatch(redirect({ type: location.type, payload }));
```
sorodrigo: is the return necessary?

Bluesmile82: Not really

PR #807, opened 2019-05-06
Fix: optimize  backgrounds

sorodrigo:
This PR optimizes some of the images used across the app and adds the ImgBackground component to entrypoints.


frontend/scripts/react-components/shared/img/img.component.jsx
```diff
@@ -29,7 +29,15 @@ function Img(props) {
         return (
           <>
             <img ref={ref} alt={alt} sizes={sizes} srcSet={srcSetProp} src={srcProp} {...rest} />
-            {!inView && <a href={src}>{alt}</a>}
+            {!inView && (
```
Bluesmile82: Shouldn't we keep the alt for SEO?

frontend/public/images/dashboards/icon_commodities.svg
```diff
@@ -1,23 +1 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
Bluesmile82: We can actually use https://github.com/tcoopman/image-webpack-loader to optimize SVGs with SVGO

frontend/scripts/react-components/shared/img/img.component.jsx
sorodrigo: it's still there in line 38 😛

Bluesmile82: Ok, monday reading 🤦‍♂ Good one!

frontend/public/images/dashboards/icon_commodities.svg
sorodrigo: Nice! Though we'd need to test it works without importing the images like
```js
import myImage from '../images'
```

sorodrigo: We already have it on prod, but not sure it's working and it's probably due to how we load the svgs. Let's review that in the future.

PR #809, opened 2019-05-07
Return title and legend_title for top sources / top destinations charts

agnessa:
https://www.pivotaltracker.com/story/show/165849709

Adds `title` and `legend_title` to `top_sources` & `top_destinations` endpoints.

Chart title is configured in charts -> title; legend title is configured in chart_attributes -> legend_name. This configuration is separate for importer & exporter profiles.


sorodrigo:
I'm receiving the title also at `chart_attributes -> title`. We should remove that one probably

agnessa:
I was under the impression that we needed both `title` and `legend_name` for one of the charts. I was partly correct, but I'm glad you raised it so I checked again! Both fields are intended to be used in case of trajectory deforestation chart on place profile, this is the difference in the database:
Screenshot 2019-05-08 at 21 15 24
🤦‍♀ 
So, it looks like perhaps indeed we don't need both, but which one to leave? I have a feeling the "display_name", or "title" is probably more widely used in other charts and perhaps it's easier to get rid of legend_name after all, as probably only the trajectory deforestation chart will be affected by that.

frontend/scripts/react-components/profiles/profile-title.component.jsx
```diff
@@ -0,0 +1,56 @@
...
+    const compiled = _template(part, { interpolate });
+    return compiled(snakeCasedParams);
+  }
+
+  if (!template) {
+    return null;
+  }
+
+  return template.split(' ').map((part, i) => {
+    const hasMustache = /^.*{{.+}}.*$/.test(part);
```
Bluesmile82: Maybe it will fail in cases where there is no space betweeen the mustached part and other ones. Or you have two strings to substitute together Eg. {{quantity}}{{unit}}
Shouldn't it be better to get everything inside the {{}} with a () in the regex?

sorodrigo: I'm using the regex only to see if the word (previously extracted using `.split(' ')`) has a mustache in it. The part of substituting is very hard so I'm just relying on lodash.

Bluesmile82: Yes, sorry I didn't explain myself. I was thinking in the case that split won't work because we have two moustaches together. Not sure if it will happen

PR #820, opened 2019-05-10
Feat: split tool reducer

sorodrigo:
This PR splits the Tool reducer into 2 reducers: toolLayers & toolLinks
Furthermore, it removes the derived state props from the reducers' state and uses selectors instead.


sorodrigo:
Ready!
There's two regressions I'm aware of. I won't fix them here, I think it makes more sense to fix them in the next PRs:

- _As a user I can select 3 nodes, deselect the second node to see the remaining nodes selected with the original selection colors unchanged._ ⚠️
- _As a user I can select a polygon in the map and if the node is visible in the sankey it will be selected in the sankey without going into expanded mode, if it’s not visible it will go into expanded mode._ ⚠️ 

Tests are failing and don't understand why, I suggest we don't waste more time on it and tackle them before merging to develop.

Bluesmile82:
_isNodeVisible function is not used in tool.actions.js But maybe you are aware

sorodrigo:
yep that's related to one of the regressions.

Bluesmile82:
I have an error when trying to go to the profiles page from a item in the legend. Maybe this is a backend error? I have some CORS problems too :/
image

Bluesmile82:
The column dropdown change is neither triggering the loading state so it could be confusing if it gets slow
image

Bluesmile82:
recolor
There are some problems with the recolour by. When you select water scarcity is loading twice. And from that moment the dropdown doesn't update the sankey state. Also some loading state would be nice. We have this error:

sankey.d3layout.js?2451:197 Uncaught TypeError: linkA.recolorBy.charCodeAt is not a function

Bluesmile82:
image
I can't see the legend info of the context layers in Paraguay soy. Not sure if its a problem that is part of this PR

PR #821, opened 2019-05-13
Feature/dashboard links

Bluesmile82:
This PR creates the links from some of the nodes in the dashboard widgets (ranking and horizontal bar chart widgets) to their correspondent profile page. 

If the meta doesn't have a profile attribute that will mean that the node doesn't have a link (eg. states, biomes, ...) and it won't be hoverable.

We still need another PR to create an SVG tooltip that will show a "Go to profile" text and the full name of the node in case it has an ellipsis

image
image


sorodrigo:
Screen Shot 2019-05-13 at 17 01 12

I wonder if it's possible to add the link to multiyear charts too. Perhaps in the legend?

frontend/scripts/react-components/shared/tooltip/tooltip.component.jsx
```diff
@@ -0,0 +1,56 @@
+import React, { useState } from 'react';
+import PropTypes from 'prop-types';
+import './tooltip.scss';
+import { Manager, Reference, Popper } from 'react-popper';
+
+function Tooltip({ reference, referenceClassName, children, type }) {
```
sorodrigo: is this file necessary?

sorodrigo:
Looks good see my comments and there's a test that needs fixing.

frontend/scripts/react-components/shared/tooltip/tooltip.component.jsx
Bluesmile82: I'm afraid yLabelsProfileInfo is not available from the backend in multiyear charts 😕

Bluesmile82:
> I wonder if it's possible to add the link to multiyear charts too. Perhaps in the legend?

I'm afraid yLabelsProfileInfo is not available from the backend in multiyear charts 😕

sorodrigo:
Ok ticket to pivotal, fix test, and lets merge.

PR #827, opened 2019-05-16
Don't show default context if choropleth is disabled for first column

Bluesmile82:
image


sorodrigo:
I was also working on a solution. 😅 You solution is bit brittle because it depends on the column position, also it only checks for first dimension. I have uploaded an alternative solution #828

PR #828, opened 2019-05-16
Prevents setting choropleth on columns that dont have it

sorodrigo:
Checks that the selected geo column has choropleth enabled before setting it.


Bluesmile82:
I feel something is still wrong: Try to change the second column:
image

sorodrigo:
Damn! It's worse than I thought. Nice catch!

sorodrigo:
Done!

PR #831, opened 2019-05-17
Fix nodename in map

Bluesmile82:
Tiny fix in profiles map
Try: http://localhost:8081/profile-actor?nodeId=30535&contextId=1&year=2017&lang=en
image


frontend/scripts/react-components/profiles/top-destinations-map.component.jsx
```diff
@@ -51,6 +51,7 @@ class TopDestinationsMap extends React.PureComponent {
 
   getTopoJsonRoot() {
     const { activeTab, countryName } = this.props;
+    console.log(activeTab, countryName, this.props);
```
sorodrigo: :eyes

Bluesmile82: 🤦‍♂

PR #833, opened 2019-05-20
Fix ranking links to profiles

sorodrigo:
This PR fixes the link to profiles in the ranking chart. It should only appear if the profile property inside `yLabelsProfileInfo` is not null.

Other miscellaneous fixes:
- Don't pass non chart data through config. I think config should only describe how the chart actually looks like, passing stuff like info or meta feels wrong.
- Translate ignored strings using imperative method: `translateText`.
- Improve _just a bit_ the logo flashing on home page (it was driving me crazy).
- Add missing total value in ranking charts and limit the number of lines the ranking chart elements occupy.


sorodrigo:
Screen Shot 2019-05-20 at 09 33 19

PR #835, opened 2019-05-20
Replace modal with non supported page for non desktop pages

Bluesmile82:
We are going to use the non-supported page for all non desktop pages

image

### DIFF
--- a/frontend/html/tool.ejs
+++ b/frontend/html/tool.ejs
@@ -10,15 +10,6 @@
       <div class="c-modal js-modal"></div>
     </div>
 
-    <div class="mobile-warning">
-      <div class="veil"></div>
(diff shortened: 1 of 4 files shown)


sorodrigo:
This should only be for the flows page. Other pages should be visible on tablet.

sorodrigo:
I think we should add a resize listener that redirects to that page when you get to the point where you can't use the sankey, also the not supported page should redirect you back when you increase the window size.

ezgif com-optimize


Another thing I noticed that might be interesting to review is that, the sankey page stops making sense around 800px width. Maybe we need to make the columns narrowers when scrolling so that it keep showing the flows until the current breakpoint. We can add it as another task.

PR #836, opened 2019-05-21
Feature/focus logistics map

Bluesmile82:
Little PR to start the Indonesia logistics map addition and take a look at the code. The map focuses depending on the commodity. This is not scalable but we are going to use it because we won't have any more countries before the redesign.

Also:

- Adds feature flag to show only Indonesia or Brazil facilities
- Adds geoJson border to map depending on feature flag


sorodrigo:
Works, it would be great to add the country borders layer the same way we're adding the bounds! 😄

PR #839, opened 2019-05-22
Feature/dynamic sitemap

Bluesmile82:
This PR creates a dynamic sitemap with sitemap_generator gem:

- Create a config script to the gem with special attention to the profiles page
- Add Capistrano task to the production file that will create/refresh the file and ping the Google and Bing servers

More information in: https://github.com/kjvarga/sitemap_generator

To check the generation of the sitemap you can use: `rake sitemap:refresh:no_ping`

I only was able to try the Capistrano script with `bundle exec cap staging deploy --dry-run` which just outputs the text of the task without running it. But it wasn't running to the end on my computer. Let me know if you find another way to try it. 

Note: We have decided not to include different languages for the moment as the current translation system won't allow the correct SEO indexing


config/sitemap.rb
```diff
@@ -0,0 +1,36 @@
...
+      :category => 'Nonprofits & Activism',
+    },
+    :changefreq => 'monthly'
+  )
+
+  add'/explore', :changefreq => 'monthly'
+  add'/flows', :changefreq => 'monthly'
+  add'/profiles', :changefreq => 'monthly'
+  contexts = Api::V3::Context.all;
+  contexts.each do |context|
```
agnessa: It is slow because it runs too many queries, there are a few things that can be improved:
1. looping over years should be inside looping over node ids, otherwise you're running the exact same queries for each year
2. nodes query needs to be filtered by context id
3. we can halve the number of queries by filtering by `profile IN ('place', 'actor')` and then dynamically inserting the current profile type into the url
4. we can use `pluck` instead of `map` to reduce the number of columns fetched from the database

So in the end we could have something like:
```
  contexts.each do |context|
    Api::V3::Readonly::Node.
      where(context_id: context.id).
      where(profile: ['place', 'actor']).
      pluck(:id, :profile).
      each do |nodeId, profile|
        context.years.each do |year|
          add "/profile-#{profile}?nodeId=#{nodeId}&year=#{year}&contextId=#{context.id}", :changefreq => 'monthly'
        end
      end
  end
```

And it’s quite quick!
Also, perhaps there's room to simplify even more and only include the last year?

config/deploy/production.rb
```diff
@@ -1,2 +1,4 @@
 server 'trase.earth', user: 'ubuntu', roles: %w{web app db}, primary: true
 set :ssh_options, forward_agent: true
+
+after 'deploy:finishing', 'sitemap:refresh'
```
agnessa: Can we add the "no ping" version to staging temporarily, so we can test in real life that the sitemap is stored where it needs to be and is publicly accessible. This might require some config changes in the web server, and it would be good to test all of that fully on staging.

Bluesmile82:
Thanks for the recommendation. It's much faster now! Let's keep the 'create' instead of 'refresh' until we know everything works fine. I created a TODO

sorodrigo:
Let's test on staging!

PR #840, opened 2019-05-23
fix onTransifexLoad not defined

sorodrigo:
There's a crash when the Transifex script loads before the JS bundle.


sorodrigo:
That's a good question, could we handle that as an improvement? or am I being lazy?

sorodrigo:
Nevermind I know what would work for sure! *goes back to coding*

PR #843, opened 2019-05-27
Feature/use dropdown in column selector

Bluesmile82:
image

The column dropdown is changed to a new simplified react component without color legend but with the same functionality and a new style variant for the dropdown. Also, the currently selected option is removed from the selectable options

AT
- [x] The user should be able to select a column item
- [x] The context should change the available options
- [x] All the subsequent actions should work the same: Sankey and map


frontend/scripts/react-components/tool/column-selector/column-selector.component.jsx
```diff
@@ -11,13 +11,11 @@ function ColumnSelector({
   handleColumnSelected
 }) {
   return (
-    <div
-      className={cx('js-dropdown c-dropdown -column-selector', {
-        '-hide-only-child': hasSingleElement
-      })}
-    >
+    <div className="js-dropdown c-dropdown">
```
sorodrigo: I think both these classes aren't necessary. And if they are it's a smell

frontend/scripts/react-components/shared/dropdown/dropdown-column.variant.scss
```diff
@@ -1,12 +1,11 @@
...
 
 .c-dropdown-component.v-column {
-  .dropdown-value {
-    padding: 0;
-    font-family: $font-family-1;
-    font-size: $font-size-regular;
-    letter-spacing: -0.5px;
-  }
+  width: 100%;
+  padding: 7px 5px;
```
sorodrigo: do we need to re-specify the padding?

frontend/scripts/react-components/tool/column-selector/column-selector.component.jsx
Bluesmile82: c-dropdown class is used for the layout right now. It will be fixed in the layout PR

PR #844, opened 2019-05-28
Feat/remove default map dimensions

sorodrigo:
This PR changes the way the app deals with default selected dimensions. Previously we imperatively set the dimensions after fetching them. Now we use a selector that declaratively does that, among other things this allows us to fetch the map nodes data in parallel to the sankey nodes and columns also making it easier to reason about.


Bluesmile82:
Water scarcity and Indigenous areas don't seem to be working
image
image

Bluesmile82:
The functionality works fine. There are some issues and improvements but they might not belong to this PR

PR #848, opened 2019-05-30
Feature/filter nodes by node type or node

agnessa:
https://www.pivotaltracker.com/story/show/166237121
https://www.pivotaltracker.com/story/show/166237113

You can now:
- `/api/v3/contexts/1/nodes` - all nodes like before
- `/api/v3/contexts/1/nodes?node_types_ids=3,4` - selected node types
- `/api/v3/contexts/1/nodes?nodes_ids=1060,1165` - selected nodes

It now loads results from a materialized view, which makes it faster in all cases.

### DIFF
--- a/app/controllers/api/v3/nodes_controller.rb
+++ b/app/controllers/api/v3/nodes_controller.rb
@@ -1,14 +1,24 @@
 module Api
   module V3
     class NodesController < ApiController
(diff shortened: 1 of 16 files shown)


sorodrigo:
🎉 🎉 🎉
 yas

PR #849, opened 2019-05-30
Feat: use sagas to load nodes columns

sorodrigo:
This PR moves some things around to start fetch data with sagas. I was able to remove an imperative call to update the fetching when context switches. Now we listen to the `SET_CONTEXT` action and fetch the data.


sorodrigo:
I'll look into it! thanks

PR #852, opened 2019-06-04
Fix some components and reorder them in menu

Bluesmile82:
Quick fix to docz components order and some bugs.

TODO:

- Accordion is not opening
- Ranking and Dynamic sentence charts need a store passed by context or direct component


sorodrigo:
Why don't we use only the components without the selectors part in the docz?

PR #854, opened 2019-06-10
Feature/logistic map indonesia

Bluesmile82:
This PR adds the new Indonesia - Palm oil context to the logistics map:

- Add a layer for Palm Oil and texts
- Add parent companies of the mills to the map
- Add some attributes to the tooltip
- Fix download modal

image

### DIFF
--- a/frontend/scripts/react-components/logistics-map/logistics-map-download/logistics-map-download.component.jsx
+++ b/frontend/scripts/react-components/logistics-map/logistics-map-download/logistics-map-download.component.jsx
@@ -7,44 +7,7 @@ import Heading from 'react-components/shared/heading/heading.component';
 
 import 'react-components/logistics-map/logistics-map-download/logistics-map-download.scss';
 
-function LogisticsMapDownload(props) {
-  const { close, layers } = props;
-  return (
-    <LogisticsMapModal
-      heading="Download"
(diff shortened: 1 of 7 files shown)


sorodrigo:
Well done! Three things to change maybe:

- In the next screenshot you can see a blank when no data is available, maybe we should show `N/A`?
Screen Shot 2019-06-12 at 17 02 41

- In the selector because only Indonesia is showing when clicking we don't see any changes, this should be disabled so that the item is read only, see the Recolor By selector on any non subnational context.

- Finally, we're reusing the crushing facilities icon for the mills, speak with Fanny to get a different icon and perhaps a different color? 🌬

Bluesmile82:
Updated the Carto layer. Still some questions to be answered. The crushing facilities icon for the mills was suggested by Pablo. Maybe @fannycc (@Fanny) has a different option

fannycc:
@Bluesmile82 I'd need more context about that icon. Which one is it and for what we are using it?

fannycc:
@Bluesmile82 @sorodrigo let's ask them about the mills layer in Indonesia to know what type of data gives. None seems to know it, not even Pablo. Can you guys ask?

sorodrigo:
well done!! 💯

PR #856, opened 2019-06-12
Feature/refactor spinner

Bluesmile82:
- Remove tool content component and move the logic to the tool component
- Move spinner to react-component

Extra:

- Fix cookie wrong proptype


frontend/scripts/react-components/tool/tool.component.jsx
```diff
@@ -199,30 +209,41 @@ const Tool = ({ resizeSankeyTool }) => {
...
+        <div
+          className={cx('flow-content', {
+            '-center-map': isMapVisible,
+            open: isVisible
+          })}
+        >
+          <div className={cx('tool-loading', { 'is-visible': loading })}>
             <div className="veil sankey-veil" />
+            <Spinner />
             <div className="c-spinner" />
```
sorodrigo: it seems you have two spinners. I'm not sure this is the best way to show the spinner, can we include the veil in the component? Also might want to use the shrinking-spinner to use the same everywhere in the app. We might want to ask a designer about the spinner/veil situation.

frontend/scripts/react-components/tool/tool.component.jsx
```diff
@@ -199,30 +209,41 @@ const Tool = ({ resizeSankeyTool }) => {
...
           </div>
 
-          <div className="js-map-view-veil sankey-veil veil is-hidden" />
+          <div className={cx('sankey-veil veil', { 'is-hidden': !isMapVisible })} />
```
sorodrigo: Ok I see why we can't use the veil within the spinner.

frontend/scripts/tests/dashboard-element/fetch.saga.test.js
```diff
@@ -1,18 +1,14 @@
 import { initialState } from 'react-components/dashboard-element/dashboard-element.reducer';
 import { cancelled } from 'redux-saga/effects';
-import { fetchWithCancel } from 'utils/saga-utils';
-
+import fetchWithCancel from 'utils/fetchWithCancel';
```
sorodrigo: I prefer having this in the other file, to fix the mocks we can do what I did on the default columns branch

PR #857, opened 2019-06-13
Feature/cms improvements

agnessa:
Improvements to managing resize by / recolor by attributes

- There's now a single entry point for both types of attributes, which is a list of contexts; it takes you to the list of resize by / recolor by scoped by context, which should make it easier than with the current long list of all attributes defined in the system
Screenshot 2019-06-13 at 11 55 07

- The list of attributes allows now for some in-place editing, without having to use the "edit" form:
    - you can toggle is_default and is_disabled
    - you can edit group number and tooltip after clicking on them
    - you can drag and drop rows, using the handles on the right, to reorder lists
Screenshot 2019-06-13 at 12 09 41

- Because attributes are now edited in scope of a context, we could remove the context dropdown from the form and the filter from the sidebar
- Position is no longer set in the form, as sorting is handled in the index page
- Available years of data are set automatically
- After creating / updating attributes using the form, it goes back straight to index


app/admin/api/v3/context.rb
```diff
@@ -16,18 +18,35 @@ def scoped_collection
     end
   end
 
+  includes [
+    :country,
+    :commodity,
+    :resize_by_attributes,
+    :recolor_by_attributes
+  ]
```
aabdaab: we can add `top_profiles` here as well

aabdaab: and I guess, as we're using `includes` now directly, we no longer need `scoped_collection`

app/models/api/v3/qual.rb
```diff
@@ -31,6 +33,13 @@ def simple_type
         'qual'
       end
 
+      def download_original_attribute(context)
+        Api::V3::DownloadQual.
+          joins(:download_attribute).
+          where('download_attributes.context_id' => context.id, qual_id: id).
```
aabdaab: maybe `find_by`? I know it works the same but it's more semantical if there's only one element like that

app/models/api/v3/quant.rb
```diff
@@ -32,6 +34,13 @@ def simple_type
         'quant'
       end
 
+      def download_original_attribute(context)
+        Api::V3::DownloadQuant.
+          joins(:download_attribute).
+          where('download_attributes.context_id' => context.id, quant_id: id).
```
aabdaab: check my comment above

app/admin/api/v3/context.rb
agnessa: excellent point, thank you!

PR #859, opened 2019-06-18
Feature/refactor tool search

Bluesmile82:
This PR refactors the search in the Sankey and map:

- Now we are using the backend search like in the homepage but filtering by context.
- Change of column and selecting a node from another column is also considered.
- Fixed some problem with the expand node on column change.

There are two cases that are still not working:

- The endpoint is returning all the years so sometimes even if we select some node it might not show in the actual year, but it will in the one that its available. We should filter the results or disable them according to the selected years

- When we select two nodes that don't have flows in common the second node won't load. 
  Try: ARAUPEL + AMINOLA. You will get the following error:
https://staging.trase.earth/api/v3/contexts/1/flows?start_year=2003&end_year=2017&include_columns=3%2C6%2C7%2C8&flow_quant=Volume&locked_nodes[]=28857&locked_nodes[]=36235&n_nodes=100&selected_nodes=28857%2C36235

In this case, we should show a banner to expand the nodes (They will then show) or to reset the Sankey options.

### DIFF
--- a/frontend/scripts/react-components/tool/tool-search/tool-search.saga.js
+++ b/frontend/scripts/react-components/tool/tool-search/tool-search.saga.js
@@ -1,31 +1,13 @@
 import { SET_SELECTED_NODES_BY_SEARCH } from 'react-components/tool/tool.actions';
 import { all, fork, takeLatest, select, put } from 'redux-saga/effects';
 import { getVisibleNodes } from 'react-components/tool/tool.selectors';
-import { expandSankey, selectColumn } from 'react-components/tool-links/tool-links.actions';
+import { expandSankey } from 'react-components/tool-links/tool-links.actions';
 
-function* selectSearchNodes() {
-  function* changeColumn(toolLinks, ids) {
-    const node = toolLinks.data.nodes[ids[0]];
-    const columnId = node?.columnId;
-    if (columnId) {
-      const nodeColumn = toolLinks.data.columns[columnId];
-      const selectedColumnsIds = toolLinks.selectedColumnsIds;
-      if (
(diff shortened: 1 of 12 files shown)


frontend/scripts/react-components/tool/tool-search/tool-search.saga.js
```diff
@@ -0,0 +1,25 @@
+import { SET_SELECTED_NODES_BY_SEARCH } from 'react-components/tool/tool.actions';
+import { all, fork, takeLatest, select, put } from 'redux-saga/effects';
+import { getVisibleNodes } from 'react-components/tool/tool.selectors';
+import { expandSankey } from 'react-components/tool-links/tool-links.actions';
+
+function* checkExpandNode() {
+  function* performCheckexpandNode({ payload }) {
+    const { results } = payload;
+    const ids = results.map(n => n.id);
+    const state = yield select();
```
sorodrigo:
```suggestion
    const visibleNodes = yield select(getVisibleNodes);
```

frontend/scripts/react-components/tool/tool-search/tool-search.saga.js
```diff
@@ -0,0 +1,25 @@
...
+    const visibleNodes = getVisibleNodes(state);
```
sorodrigo:
```suggestion

```

sorodrigo:
When selecting a node via search the linked geo ids should recalculate too. This can be done by subscribing the `fetchLinkedGeoIds` saga to `SET_SELECTED_NODES_BY_SEARCH`

Bluesmile82:
Ready again

PR #860, opened 2019-06-19
Allow managing Top Profiles in the CMS

aabdaab:
PT task: https://www.pivotaltracker.com/n/projects/1637931/stories/166730157

* Add top_profiles table to the DB: associations with context and node
* Add associations to context and node models
* Add CMS pages for context and top profiles
* Filter available nodes for top profiles

1. New `Contexts` tab in the CMS (it's not nested)
2. Index page show dot-listed contexts with links to the details about a single context (looks like `Dashboard` index)
3. Context details include a link to `Top Profiles`
4. Top Profiles are listed for that specific context
5. Add page for Top Profile is filtering the available nodes (these which have profile pages defined in that context)
6. I used materialized views, turned off the filters in the CMS and used scoped collection in the `Context` for performance reasons.

**Note**:
This PR is missing specs but I thought I'm gonna open it anyway in case there are some business requirements I missed that can be caught on this stage.

### DIFF
--- a/app/admin/api/v3/context.rb
+++ b/app/admin/api/v3/context.rb
@@ -1,0 +1,42 @@
+ActiveAdmin.register Api::V3::Context, as: 'Context' do
+  menu priority: 2
+
+  config.filters = false
+
+  after_action :clear_cache, only: [:create, :update, :destroy]
+
+  controller do
+    def clear_cache
+      clear_cache_for_regexp('/api/v3/contexts')
+    end
+
+    def scoped_collection
+      super.includes(:country)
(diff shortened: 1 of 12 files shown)


aabdaab:
do I need to do something specific for the travis not to fail? I ran the migrations for the test environment locally but I'm having the following error:
image

agnessa:
for the test env, try doing it like this (on a clean structure.sql)
`RAILS_ENV=test rake db:drop db:create db:structure:load db:migrate`

aabdaab:
hmm, getting a different error now:
image

PR #862, opened 2019-06-19
Endpoint for top profiles

aabdaab:
PT task: https://www.pivotaltracker.com/n/projects/1637931/stories/166695827
This PR adds an endpoint for top profiles for specific context.


app/controllers/api/v3/top_profiles_controller.rb
```diff
@@ -0,0 +1,12 @@
+module Api
+  module V3
+    class TopProfilesController < ApiController
```
agnessa: we need a `skip_before_action :load_context` here, as this endpoint does not have a mandatory context paramete

app/controllers/api/v3/top_profiles_controller.rb
```diff
@@ -0,0 +1,12 @@
...
+      def index
+        @top_profiles = Api::V3::TopProfile.where(context_id: params[:context_id])
```
agnessa: let's not make context_id a required param

app/serializers/api/v3/top_profiles_serializer.rb
```diff
@@ -0,0 +1,32 @@
+module Api
+  module V3
+    class TopProfilesSerializer < ActiveModel::Serializer
```
agnessa: @Bluesmile82 or @sorodrigo can you take a look at this serializer and let Marta know in case you need any more data for the top profiles feed? Also if you need any filters on this endpoint.

app/serializers/api/v3/top_profiles_serializer.rb
```diff
@@ -0,0 +1,32 @@
...
+
+      def profile_type
+        object.node.node_type.context_node_types.find_by(context_id: object.context_id).profile.name
+      end
+
+      def year
+        object.context.years.max
+      end
+
+      def summary
```
agnessa: yes, this is exactly what we should do! I'm just worried it might be slow, because the top profiles will usually have a lot of data to process. can you test by setting some of the bigger ones as top profiles, look at the top nodes on the sankey in the first and the third column for Brazil-soy, Paraguay-soy and Paraguay-beef, set ~20 top profiles and see how long it takes.

aabdaab: You're right, it doesn't look so good; for 9 bigger nodes it takes to load up to ~13 seconds. For comparison, without the summary it's 78.87 ms..

app/controllers/api/v3/top_profiles_controller.rb
aabdaab: 👍

app/controllers/api/v3/top_profiles_controller.rb
aabdaab: 👍

app/serializers/api/v3/top_profiles_serializer.rb
aabdaab: 👍

app/serializers/api/v3/top_profiles_serializer.rb
aabdaab: maybe we could add `summary` to the top profiles table, that should make it quicker. wdyt @agnessa ?

agnessa: sure! you could also add the year, so that it's clear from which data it was generated

aabdaab:
☝️ ready again

PR #885, opened 2019-07-01
Clear cache after changes for top profiles

aabdaab:
Clear cache after changes for top profiles on _update_, _create_ and _destroy_

### DIFF
--- a/app/admin/api/v3/top_profile.rb
+++ b/app/admin/api/v3/top_profile.rb
@@ -11,7 +11,7 @@ ActiveAdmin.register Api::V3::TopProfile, as: 'Top Profile' do
     def clear_cache
       clear_cache_for_regexp('/api/v3/contexts')
     end
-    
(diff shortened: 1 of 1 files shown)


app/admin/api/v3/top_profile.rb
```diff
@@ -5,8 +5,13 @@
 
   # before creating add summary, year and profile_type to top profile record
   before_create :derive_top_profile_details
+  after_action :clear_cache, only: [:create, :update, :destroy]
 
   controller do
+    def clear_cache
+      clear_cache_for_regexp('/api/v3/contexts')
```
agnessa: I think the url in this case is different: `/api/v3/top_profiles`?

aabdaab: 🤦‍♀

aabdaab:
fixed

PR #887, opened 2019-07-03
Top Profiles: Return last available year of data per node

aabdaab:
PT task: https://www.pivotaltracker.com/story/show/167063535
Top Profiles: Return last available year of data per node and not per context. Use `nodes_mv` for that purpose.

### DIFF
--- a/app/services/api/v3/top_profiles/derive_top_profile_details.rb
+++ b/app/services/api/v3/top_profiles/derive_top_profile_details.rb
@@ -1,0 +1,42 @@
+module Api
+  module V3
+    module TopProfiles
+      class DeriveTopProfileDetails
+        attr_reader :top_profile
+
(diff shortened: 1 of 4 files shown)


app/services/api/v3/top_profiles/derive_top_profile_details.rb
```diff
@@ -0,0 +1,56 @@
...
+            top_profile.context, node, year
+          ).call[:summary]
+        end
+
+        def node
+          top_profile.node
+        end
+
+        def profile_type
+          node.node_type.context_node_types.find_by(context_id: top_profile.context_id).profile.name
```
agnessa: you could get the profile type from nodes_mv as well

aabdaab:
ready ☝️

PR #888, opened 2019-07-03
Top profiles images pool

aabdaab:
PT task: https://www.pivotaltracker.com/story/show/166790170

1. Add new tab to the menu: `Top Profile Images` where admin can upload images (without assigning them at this point)
image
2. Create new top profile image and add it to a certain pool: assign it to specific `commodity` and `profile_type`; it means that when assigning images to top profiles we will filter the available images by top profile's context's commodity and its profile type (think: tags)
image
3. After successful record creation redirect to an index page of top profile images. Filter by `commodity` and `profile_type` on the right side.
image
image
4. Assign top profile image to a top profile;
- first create a top profile
- after successful creation it redirects you to an edit page where you can assign an image from the available ones
image
It's done on radio buttons, the UI could be improved with some jquery probably.


aabdaab:
Thanks Agnieszka! Will take care of it now.

aabdaab:
I followed this advice to register the same class twice in active admin: https://github.com/activeadmin/activeadmin/issues/2306#issuecomment-23500610
This allows us to have `Contexts` both in `General` tab and in `Profiles` tab under a different name: `Top Profiles For Contexts`.
I removed the links to `Top profiles` from the `Contexts` tab and in `Top Profiles For Contexts` I'm displaying only these contexts that have profile pages.
`Top Profile Images` are under `Profiles` tab as well.

aabdaab:
ready ☝️

PR #889, opened 2019-07-04
Add list of available years for node in profile metadata endpoint

aabdaab:
Add list of available years for node in profile metadata endpoint.

### DIFF
--- a/app/controllers/api/v3/profile_metadata_controller.rb
+++ b/app/controllers/api/v3/profile_metadata_controller.rb
@@ -8,11 +8,24 @@ module Api
           @context, @node
         ).call
 
-        render json: @result,
-               root: 'data',
-               serializer: Api::V3::ProfileMetadata::ProfileSerializer,
-               include: {charts: :children},
-               key_transform: :underscore
+        serialized_profile_metadata =
+          ActiveModelSerializers::SerializableResource.new(
+            @result,
+            serializer: Api::V3::ProfileMetadata::ProfileSerializer,
+            key_transform: :underscore,
+            root: 'data',
+            include: {charts: :children}
+          ).serializable_hash
+
+        serialized_available_years =
+          ActiveModelSerializers::SerializableResource.new(
+            @node,
+            root: 'data',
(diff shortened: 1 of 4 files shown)


aabdaab:
i keep getting uncomunnicative module name issue on codeclimate for the `V3` module; should I change the `config.reek`? 
image
although how was it passing until now?

agnessa:
> i keep getting uncomunnicative module name issue on codeclimate for the `V3` module; should I change the `config.reek`?
> image
> although how was it passing until now?

hi, yes please change it if you know how, I tried before and it keeps coming up 😓 The way it was passing was because whenever I review the code climate issues on a PR, I set those ones to "wontifix" manually 😄

agnessa:
About the records not present in nodes_mv: that should not be normally possible if you go to a profile page through the profile search (because it feeds from the same mview). In other words, all nodes with profiles are supposed to be included in that mview, so let's check exactly what is going on with the particular nodes that give you the error.

aabdaab:
as for the codeclimate issue, I would gladly do that, but I don't have the option visible :D can you grant me some permissions?

aabdaab:
for example node is not found in the materialized view for this case:
http://localhost:3000/api/v3/contexts/1/nodes/4366/profile_metadata
node with id `4366`

agnessa:
> for example node is not found in the materialized view for this case:
> http://localhost:3000/api/v3/contexts/1/nodes/4366/profile_metadata
> node with id `4366`

That's municipality "WANDERLEY", the reason it is not in the materialised view is because it has no flow data. How did you arrive at this node? You wouldn't get there through the sankey or through the profile search I think, which is probably why we never had profile issues with these kinds of nodes so far.

aabdaab:
Then it makes sense. I was randomly choosing nodes, I didn't arrive there nor from the profile search nor the sankey so I think it's safe to leave it this way. I added a `rescue` anyway so we shouldn't get a 404 in any case.

aabdaab:
ready ☝️

PR #893, opened 2019-07-09
Feature/profiles sources panel

Bluesmile82:
This PR adds the commodites and sources panel. Check this one instead of ```feature/profiles-sources-commodity-panel```

- Added Step, Sources and Commodities panel
- Added footer buttons on the bottom disabling them when needed and dynamic sentence
- Sagas to fetch data and search

Note: Imported countries and countries selection in sources profile is disabled as the countries profiles are not ready yet.

Go to profile action is still not implemented

Known bug: 🐞 if you select Brazil and then Paraguay the items of the first tab are not changing

image

image
image


sorodrigo:
Some initial comments, I haven't gone through most of the code:

We probably want to center this.
Screen Shot 2019-07-10 at 13 54 23

Profile pages used to be mobile-friendly, we need to figure out what to do here.
Screen Shot 2019-07-10 at 13 54 04

We need to filter out, the tabs and gridlist items (in countries section) that don't have profiles.
Screen Shot 2019-07-10 at 12 46 11

Screen Shot 2019-07-10 at 13 59 46

Perhaps we should show the commodity in the dynamic sentence?
Screen Shot 2019-07-10 at 14 00 31

Bluesmile82:
- Not showing the commodity in the dynamic sentence was a design decision as it could confuse the user.
- About the responsive, I agree. This will help with the dashboards responsive too.
- Maybe some attribute from the backend would be nice for the tab filtering as we are trying to get rid of hardcoded filters

agnessa:
> * Maybe some attribute from the backend would be nice for the tab filtering as we are trying to get rid of hardcoded filters
for the tabs: https://github.com/Vizzuality/trase/pull/896

PR #894, opened 2019-07-09
Feature/enable extra indicators on dashboards

agnessa:
New indicators:
* Environmental embargoes
* Slave labour
* Unprotected vegetation in private lands
* Legal reserve deficit

1. Need to be enabled in the admin panel (http://localhost:3000/content/admin/dashboards_attributes) 
2. Data for these indicators is per-node, available only for some nodes (typically sourcing regions). If a sourcing region is part of the selection, extra charts are returned to show values of these indicators.
3. In case of single year selection for a temporal indicator, a dynamic sentence is returned per indicator per node and `meta` -> `info` -> `temporal` is **true**
`meta` -> `info:`
```
      "info":{  
         "temporal":true,
         "years":{  
            "start_year":2015,
            "end_year":null
         },
         "filter":{  
            "cont_attribute":"Reported cases of forced labour",
            "node":{  
               "id":10902,
               "name":"SORRISO",
               "node_type_id":3,
               "node_type":"MUNICIPALITY",
               "profile":"place"
            }
         }
      }
```
4. In case of single year selection for a non-temporal indicator, a dynamic sentence is returned per indicator per node and `meta` -> `info` -> `temporal` is **false**
4. In case of multi year selection for a temporal indicator, a bar chart is returned per indicator per node and `meta` -> `info` -> `temporal` is **true**
5. In case of multi year selection for a non-temporal indicator, a **dynamic sentence** is returned per indicator per node and `meta` -> `info` -> `temporal` is **false**

Current issues:

- some of these indicators are non-temporal, meaning value is not for a given year but "in general for a node". In such cases a dynamic sentence is returned; however, it should not display the year / year range in such case, for example "Unprotected vegetation in private lands"
Screenshot 2019-07-09 at 16 07 16

- the value is per node and does not depend on other selected nodes or even context, so probably we should only give the name of the indicator, the value + year (where temporal) and omit all else
Screenshot 2019-07-09 at 16 11 19

- some units are better than others...
Screenshot 2019-07-09 at 16 12 06


sorodrigo:
@agnessa I made the necessary changes to sentence and titles (still awaiting for dropdown so it will look better then). I found an issue:
This is rendered in the *Paraguay Beef plus 1+ Departments* scenario.

Screen Shot 2019-07-11 at 15 06 42

agnessa:
I added a db migration to remove any previously defined dashboard attributes and create the 4. Also added chart grouping.

PR #897, opened 2019-07-10
Feature/profiles companies panel

Bluesmile82:
This PR:

- Creates the companies panel with the countries dropdown
- Creates dynamic sentence on the footer
- Link to profiles (The year will come from the backend so is not ready yet)

Missing:

- Show only Countries with sources and traders data - Countries will be filtered by the backend
- [x] Solve Countries changing bug
- [x] Fix tabs filtering
- Responsive - needs design

image
image


sorodrigo:
- Once I select and item on the gridlist I'm not able to deselect it.
- Pagination is not working on the companies section.
- When navigating to profile page, the modal state is not being cleared. I think this state should live at the component level instead of redux.
- When navigating to profile, the browser's back button stops working.
- I can't access the brazil/beef profile pages.

Screen Shot 2019-07-15 at 12 39 26
Screen Shot 2019-07-15 at 12 41 47

Bluesmile82:
Thank you for the comments:

- We talked with design and decided we won't deselect or clear items as we are only able to select one.
- I will check the pagination and the back button. Maybe something is wrong in the redirection.
- About the modal state I wanted to make it work for the URL but maybe you are right and we should keep it in the state.
- Brazil Beef for municipalities is also not accessible on dashboards. I'm not sure if we have the data for this or its a bug. @agnessa do you have more info?
- I should have noted that backend will provide the last year for each node. That is why now its NaN

Bluesmile82:
Ready again!
- I added the change profile button on the profile page. The places page needs some design changes that will be done in a different PR. 

Also:
- Profiles are filtered by the backend now.
- The modal state it's still in redux because is needed in the sagas. The modal step resets when we close it when we go to the profile node.

frontend/scripts/react-components/shared/profile-selector/profile-panel/profile-panel.saga.js
```diff
@@ -0,0 +1,207 @@
...
+    yield fork(getProfilesData, 'countries');
+    // Fetch regions
+    if (!isEmpty(profileSelector.panels.countries.activeItems)) {
+      yield fork(getProfilesData, 'sources');
+    }
+  } else if (panelName === 'companies') {
+    yield call(getProfilesData, 'countries');
+    const updatedProfileSelector = yield select(state => state.profileSelector);
+    if (isEmpty(updatedProfileSelector.panels.countries.activeItems)) {
+      const defaultCountry = updatedProfileSelector.data.countries.find(
```
sorodrigo: something could happen in indonesiademo when we want to add profiles

frontend/scripts/react-components/shared/profile-selector/profile-selector.actions.js
```diff
@@ -109,10 +74,61 @@ export const setProfilesLoadingItems = loadingItems => ({
     loadingItems
   }
 });
+export const goToProfile = () => (dispatch, getState) => {
+  const { profileSelector, app } = getState();
+  const { contexts } = app;
+  const hasCompanies = !isEmpty(profileSelector.panels.companies.activeItems);
+  const profileType = hasCompanies ? 'actor' : 'place';
+  const profileSelection = hasCompanies
+    ? Object.values(profileSelector.panels.companies.activeItems)[0]
```
sorodrigo: why not use a activeItem instead of an array

frontend/scripts/react-components/shared/profile-selector/profile-selector.actions.js
```diff
@@ -109,10 +74,61 @@ export const setProfilesLoadingItems = loadingItems => ({
...
+    : Object.values(profileSelector.panels.sources.activeItems)[0];
+  const query = {
+    nodeId: profileSelection.id
+  };
+  const commodity = Object.values(profileSelector.panels.commodities.activeItems)[0];
```
sorodrigo: we dont need commodity

frontend/scripts/react-components/shared/profile-selector/profile-selector.selectors.js
```diff
@@ -0,0 +1,50 @@
...
+    if (!isEmpty(panels.sources.activeItems)) {
+      dynamicParts.push({
+        panel: 'sources',
+        id: 'sources',
+        prefix: 'See the',
+        value: Object.values(panels.sources.activeItems)
+      });
+    } else if (!isEmpty(panels.companies.activeItems)) {
+      dynamicParts.push({
+        panel: 'sources',
```
sorodrigo: this should be companies

frontend/scripts/react-components/shared/tabs/tabs.component.jsx
```diff
@@ -14,35 +14,39 @@ function Tabs(props) {
...
+                weight="bold"
+                size="rg"
+                variant="mono"
+              >
+                {itemTabRenderer ? itemTabRenderer(item, index) : item}
+              </Text>
+            </button>
+          ))}
+        </div>
+        {extraDropdown}
```
sorodrigo: I don't think this should be part of the tabs. The dropdown is a separate component that should be a sibling.

frontend/scripts/react-components/shared/tags-group/tags-group.component.jsx
```diff
@@ -7,29 +7,43 @@ import { translateText } from 'utils/transifex';
...
+          align="center"
+          key={part.id}
+          color={color}
+          className="tag-group-part notranslate"
+        >
+          {part.prefix && translateText(`${part.prefix} `)}
+          {part.value && <Tag {...props} part={part} />}
+        </Heading>
+      ))}
+      {suffix && tags.length > 0 && (
```
sorodrigo: It's confusing to have `part.prefix` and `suffix`. I think this could be achieved by adding a new part that only has a value `profile`

frontend/scripts/react-components/shared/profile-selector/profile-selector.actions.js
Bluesmile82: This is to get the contextId it will change depending on the context

sorodrigo: I see, my bad

frontend/scripts/react-components/shared/profile-selector/profile-selector.actions.js
Bluesmile82: I tried but we are reusing the panels from dashboards and it would be even more confusing there

Bluesmile82:
Ready again

frontend/scripts/react-components/shared/profile-selector/profile-selector.actions.js
sorodrigo: I see, you're right. But then the CompaniesPanel prop should be `activeCompanies` instead of `activeCompany`

PR #901, opened 2019-07-11
Use flexbox to mantain the whole title always visible

Bluesmile82:
This PR refactors AnimatedCard component to use flexbox and avoid having a min-height
In this way, we can always see the full title of the profile no matter how long it is.

image
image


Bluesmile82:
Test fixed. Please review

sorodrigo:
ScreenRecording2019-08-20at11590
With Flexbox the animation gets a bit jumpy. I spoke with Pablo and he prefers having the cards with the same title height.

PR #905, opened 2019-07-11
Get random matching img for top profile

aabdaab:
PT task: https://www.pivotaltracker.com/story/show/167162476

Added a service `GetMatchingImage` that first tries to find matching top profile images (same **commodity** and **profile_type**) - if it doesn't find anything, returns `nil`. We could have some predefined images for that if it's not a desired behaviour. Let me know in the comment.

If it has some matching images then it tries to find the ones that are not assigned to any top profile; if it does, it returns a random one, if not, it returns a random one from the matching images.

Everything happens on the fly, so there are no assignments in the DB. If we want to persist that, let me know in the comments as well.

### DIFF
--- a/app/serializers/api/v3/top_profiles_serializer.rb
+++ b/app/serializers/api/v3/top_profiles_serializer.rb
@@ -15,7 +15,12 @@ module Api
       end
 
       def photo_url
-        object&.top_profile_image&.image&.url
+        if object.top_profile_image
+          object.top_profile_image.image.url
+        else
+          top_profile = Api::V3::TopProfile.includes(:context).find(object.id)
+          Api::V3::TopProfiles::GetMatchingImage.call(top_profile)
+        end
       end
     end
   end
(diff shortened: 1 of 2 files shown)


app/serializers/api/v3/top_profiles_serializer.rb
```diff
@@ -15,7 +15,12 @@ def node_type
       end
 
       def photo_url
-        object&.top_profile_image&.image&.url
+        if object.top_profile_image
+          object.top_profile_image.image.url
+        else
+          top_profile = Api::V3::TopProfile.includes(:context).find(object.id)
```
agnessa: why do we need this? isn't object == top_profile?

aabdaab: it is, but it's an instance and I wanted to avoid n+1 queries (i'm getting info from the `context` in the service)

PR #907, opened 2019-07-12
When in comparison view return bar_chart type

aabdaab:
PT task: https://www.pivotaltracker.com/story/show/167012059

When there are more than 1 node in the dashboard panel selected (comparison view) return `BAR_CHART` type.

### DIFF
--- a/app/services/api/v3/dashboards/parametrised_charts/flow_values_charts.rb
+++ b/app/services/api/v3/dashboards/parametrised_charts/flow_values_charts.rb
@@ -60,6 +60,14 @@ module Api
             )
           end
 
+          def comparison_view?
+            nodes.size > 1
+          end
+
+          def get_correct_chart_type(default_chart_type)
+            comparison_view? ? BAR_CHART : default_chart_type
+          end
+
           def initialize_node_types
             node_types_to_break_by =
               NodeTypesToBreakBy.new(@context, nodes)
@@ -87,7 +95,7 @@ module Api
           def single_year_no_ncont_overview
             {
               source: :single_year_no_ncont_overview,
-              type: DYNAMIC_SENTENCE,
+              type: get_correct_chart_type(DYNAMIC_SENTENCE),
               x: nil
             }
           end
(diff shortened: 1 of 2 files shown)


app/services/api/v3/dashboards/parametrised_charts/flow_values_charts.rb
```diff
@@ -87,23 +95,23 @@ def single_year_no_ncont_charts
           def single_year_no_ncont_overview
             {
               source: :single_year_no_ncont_overview,
-              type: DYNAMIC_SENTENCE,
+              type: get_correct_chart_type(DYNAMIC_SENTENCE),
```
agnessa: in general let's not apply this logic to any "overview" charts, only "node type view" ones

app/services/api/v3/dashboards/parametrised_charts/flow_values_charts.rb
```diff
@@ -60,6 +60,14 @@ def nodes
             )
           end
 
+          def comparison_view?
+            nodes.size > 1
+          end
+
+          def get_correct_chart_type(default_chart_type)
```
agnessa: I think this method should be taking the node type as parameter, and the comparison_view? check must only take into account nodes of the same node type per chart, and not globally (e.g. if 2 municipalities are selected, the "monicipality node type view" is a comparison view, but the "exporter node type view" is not

aabdaab: just to double-check, when you say `selected` you mean, predefined by backend and not selected on the client? so, when we've got more than 1 records in here, i.e.: 
image
this is a comparison view?
Also, I noticed I'm returning a `BAR_CHART` type for these cases, to make sure: shouldn't it be a `RANKING_CHART`?

agnessa:
Maybe it is easier to see on this branch, because it has more charts enabled:
https://github.com/Vizzuality/trase/pull/898

By selected I mean "user selected"

Example: user selects 2 sourcing municipalities (Brasnorte and Campos de Julio) and 2 exporters (Amaggi and Bunge). For the "single year no ncont" scenario we get 2 comparison charts: one for municipalities and one for exporters.

What I get is:
exporters comparison is a bar chart, which is correct:
single_year_no_ncont_two_exporters_node_type_view

municipalities comparison is a ranking chart, and should be a bar chart:
single_year_no_ncont_two_exporters_two_sources_node_type_view_1

I think this particular issue only applies to single year + no ncont scenario. There's a slightly different issue for multi year scenarios, that's a separate task. So ideally we should see no change to any other charts but the "node type views" for selected node types in single year no-ncont scenario 😅

aabdaab:
^ applied changes

agnessa:
Awesome, thanks!

Issue #717, opened 2019-02-25
Create project domain documentation


agnessa:
We need an introduction into the concepts in Trase. This is the proposed minimum table of contents:

Introduction to core concepts
- contexts
- nodes and types of nodes
- flows and flow paths
- inds, quants and quals
- node attributes and flow attributes
Core data vs configuration
- blue vs yellow tables
- description of currently available configuration options

aabdaab:
I think this could be a good starting point. I would much benefit from docs about the topics above. Even things like we discussed in here #712, like:
What is a flow? What compounds make a flow, from that we can draw what are the limitations of a flow, i.. there can be only one country of destination in a single flow (single source as well then?)
Or, what you said in PT task: `It should never happen that 2 companies of same "type", i.e. 2 exporters, will appear at the same time in the path`.
We could start with a glossary of terms we use in a project on a daily basic (flow, path, context, ...) and basic use cases. I think from that it will start to grow because once we start listing the basics, more complex scenarios will come to mind (edge cases, implications on a different part of application, things like that).

agnessa:
https://github.com/Vizzuality/trase/pull/909

PR #909, opened 2019-07-15
Onboarding documentation start

agnessa:
This is the beginning of backend onboarding documentation, based on what we covered with Jesús so far.

Here are some comments from Marta explaining the need of such documentation: https://github.com/Vizzuality/trase/issues/717

I think it would be great if Jesús would add either information or questions here as he's now discovering Trase, then hopefully the next person who starts will have an easier job 🌻


jesusgonzalezrivera:
I think we should add documentation about how to access to the administration page. For example, I had tried to access to the administration page on staging but I don't know the user accounts.

agnessa:
> I think we should add documentation about how to access to the administration page. For example, I had tried to access to the administration page on staging but I don't know the user accounts.

Sure, we use individual accounts to the admin panel, so you need to ask someone to create an account for you.

PR #911, opened 2019-07-15
In comparison view dont include other in legend

aabdaab:
PT task: https://www.pivotaltracker.com/story/show/167065042
For selected nodes don't include `OTHER` in the legend.

Before:
image

After:
image

Note that on the attached second screen we're displaying `OTHER` in the legend of `Top 10 Importers` because no importer was selected before.

To test this scenario:
First pick `Brazil`, then from optional regions: `Brasnorte` and `Campos de Julio` --> `SOY` --> nothing --> from exporters tab `Bunge` and `Amaggi`


agnessa:
> Note that on the attached second screen we're displaying `OTHER` in the legend of `Top 10 Importers` because no importer was selected before.

yes, that's perfect, thank you!

PR #914, opened 2019-07-16
Use brazil_storage_facilities instead of the sample dataset

Bluesmile82:
Change the dataset for brazil:

Before:
image

After:
image


sorodrigo:
working, but some backend tests fail 🤔

PR #915, opened 2019-07-16
fix for top profile images

agnessa:
I know this looks weird, but it is effectively the same "design pattern" as in https://github.com/Vizzuality/trase/blob/develop/app/serializers/content/post_serializer.rb and it makes it work on staging ¯\_(ツ)_/¯

### DIFF
--- a/app/admin/api/v3/top_profile.rb
+++ b/app/admin/api/v3/top_profile.rb
@@ -33,7 +33,7 @@ ActiveAdmin.register Api::V3::TopProfile, as: 'Top Profile' do
       if property.top_profile_image.nil?
         'No image assigned'
       else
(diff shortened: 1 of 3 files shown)


aabdaab:
I approved it but now I'm not so sure: actually the images are working on frontend, see here: https://staging.trase.earth/profiles
and the change you added is in the serializer without which images are displaying on the frontend, just not on the admin panel.

agnessa:
Thanks! The reason they were displaying on staging is because I "hot fixed" staging to check this. Yesterday I freshly deployed develop to staging, and they're not working again :) But thank you for pointing out about the admin panel, you're right that still needs fixing.

PR #919, opened 2019-07-17
Feature/profiles place summary redesign

Bluesmile82:
This PR branches from ```feature/profiles-companies-panel```

- Implement the new design for the municipality profiles
- Use Text and Icon components for the text and standardize it with the actor profiles
- Change button to a more slim version (xs size)
- Extract title part into SummaryTitle component

EXTRA: 

- This PR also fix the problem with the years in profiles. Only including in the dropdown the available years for each node and not sending year param on any link to profiles

image

image


Bluesmile82:
Tests fixed. Ready to go!

sorodrigo:
Good job! However I think we need to revisit the profile-place responsive designs. And we should hide the browse button on mobile.
*production*
Screen Shot 2019-07-17 at 19 15 20

*now*

Screen Shot 2019-07-17 at 19 18 12
Screen Shot 2019-07-17 at 19 15 18

sorodrigo:
Also in the previous comment where the `Area` and `Soy Land` values are displayed. The font size is too large. In profile place is set as `lg` and in profile actor is set as `rg`

Bluesmile82:
I don't think is too large. Its size according to the rest of the text. Maybe we can enlarge the profile actor text. This was the blueprint from design:
image

PR #922, opened 2019-07-19
Feat: chart widget groupings

sorodrigo:
Adds support for groupings in dashboard widgets, this allows backend to specify grouping of charts that will be shown using a dropdown.


sorodrigo:
ready 👍

Bluesmile82:
Please check the failing tests. Maybe we should deactivate codeclimate. Is distracting

PR #923, opened 2019-07-19
Feature/basic attributes

jesusgonzalezrivera:


app/services/api/v3/actors/basic_attributes.rb
```diff
@@ -57,19 +62,34 @@ def summary
...
+            chart_attribute = @chart_config.named_chart_attribute(name)
+            values[:header_attributes][chart_attribute.identifier.to_sym] =
+              header_attributes(chart_attribute)
+          end
+          values
+        end
+
+        def header_attributes(attribute)
+          {
+            name: attribute.display_name,
```
agnessa: the name comes up empty for me for some reason

Screenshot 2019-07-19 at 22 40 19

should be "Forest 500 Score" and "Zero Deforestation Commitment" respectively I think

agnessa:
FYI @sorodrigo @Bluesmile82 
We have a new way of returning data for the "header attributes" on profile pages (in basic_attributes endpoints); right now we serve the values "the old way" and "the new way" simultaneously so that you don't need to amend front end immediately. This will allow us to remove the hardcoded "labels" and "tooltips" we're currently using on the frontend, since those are coming from the backend now.

For the place profile:
Screenshot 2019-08-02 at 14 31 45

For the actor profile:
Screenshot 2019-08-02 at 14 34 52

PR #924, opened 2019-07-19
Disable interaction for mobile in explore

Bluesmile82:
Disable tooltips for mobile in explore component. This was made for mobile UX reasons, the size of the action tooltips was too little. Maybe in the future would be nice to have some list with the countries

Also, checking the size of the breakpoint I realise that this doesn't fit most of the mobile width. We shall review this breakpoints.

image

### DIFF
--- a/frontend/scripts/utils/isMobile.js
+++ b/frontend/scripts/utils/isMobile.js
@@ -1,0 +1,3 @@
+import { BREAKPOINTS } from 'constants';
+
+export default window.innerWidth <= BREAKPOINTS.small;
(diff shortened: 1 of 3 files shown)


frontend/scripts/utils/isMobile.js
```diff
@@ -0,0 +1,3 @@
+import { BREAKPOINTS } from 'constants';
+
+export default window.innerWidth <= BREAKPOINTS.small;
```
sorodrigo: I would make this a function just so it gets evaluated every time it's used.

PR #930, opened 2019-07-29
Fix/improve geo columns loading

Bluesmile82:
This PR improves geo columns loading using 'fork' instead of 'cal'l for parallel loading. 

Also:
- Removes some unnecessary geoId logic as we have it now in the contexts.
- Don't render until we have the nodes for the four columns
- Merges getToolNodes with getMoreToolNodes: The logic in getMoreToolNodes was already taking into account the first case and we needed some checking for duplicates in the first case

### DIFF
--- a/frontend/scripts/react-components/tool-links/tool-links.reducer.js
+++ b/frontend/scripts/react-components/tool-links/tool-links.reducer.js
@@ -34,13 +34,13 @@ import * as ToolLinksUrlPropHandlers from 'react-components/tool-links/tool-link
 import toolLinksInitialState from './tool-links.initial-state';
 
 function setNodes(state, action) {
-  const { nodes } = action.payload;
+  const { nodes, replaceData } = action.payload;
   return immer(state, draft => {
     nodes.forEach(node => {
-      if (!draft.data.nodes) {
+      if (!draft.data.nodes || replaceData) {
         draft.data.nodes = {};
       }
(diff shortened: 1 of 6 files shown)


Bluesmile82:
Ready again!

frontend/scripts/react-components/tool-links/tool-links.reducer.js
```diff
@@ -142,22 +141,9 @@ const toolLinksReducer = {
       // TODO: if any selectedNode, make those columns visible (selected)
     });
   },
+  [TOOL_LINKS__SET_NODES]: setNodes,
 
-  [TOOL_LINKS__SET_NODES](state, action) {
-    const { nodes } = action.payload;
-    return immer(state, draft => {
-      draft.data.nodes = {};
```
sorodrigo: My only concern is that this gets triggered everytime we change context, and if we use only one action we need a way of deleting the existing nodes.

Bluesmile82: Yeah, good one. But shouldn't we store the links and nodes by context so we don't have to retrieve them again every time we change the context?

sorodrigo: The browser will cache them, so we don't need it. And that way we avoid memory leaks.

Bluesmile82:
Ready again

PR #933, opened 2019-07-31
Feature/flow response

jesusgonzalezrivera:
https://www.pivotaltracker.com/story/show/166237087


agnessa:
@sorodrigo remember when we talked about removing the quant / qual / ind from flows response, which could be safely done as it is not used in the frontend? Apparently it is? 😅 

Screenshot 2019-08-02 at 13 55 01

sorodrigo:
Ok I can get it from the resize by list, so it should be fine will push a fix to this branch

sorodrigo:
I made the change so now it will use the resizeBy name, something I missed is the fact that the resizeby items don't include the unit. Could we make a change so that the resize by also includes unit?

PR #935, opened 2019-08-01
Fix/optimize explore map

Bluesmile82:
This PR refactors and optimizes the explore map. Changes the logic to hooks and uses some memoization

The disableOptimization prop can't be removed as we are changing the colors of the geographies with the props

performant


sorodrigo:
great job 👏

PR #941, opened 2019-08-08
Feature/context node type properties prefix

jesusgonzalezrivera:
Add prefix field to context node type properties to set the linked text between nodes. There is a task which permits to set the default prefixes for the properties depending of the node type name:
```
RAILS_ENV={{env}} bundle exec rake context_node_type_properties:add_prefixes
```


app/models/api/v3/context_node_type_property.rb
```diff
@@ -42,6 +43,7 @@ class ContextNodeTypeProperty < YellowTable
 
       belongs_to :context_node_type
 
+      validates :prefix, presence: true
```
agnessa: I think we could relax that to only when "role" is present (otherwise node types do not appear as tabs in the dashboards filter panel)

PR #947, opened 2019-08-13
Feature/commodities improve

jesusgonzalezrivera:


config/routes.rb
```diff
@@ -80,6 +80,7 @@
         resources :commodities, only: [:index] do
           get :search, on: :collection
         end
+        resources :commodities_from_contexts, only: [:index]
```
agnessa: would it be possible to take it out of the /dashboards namespace and just call it 'commodities'? It is not dashboards specific, and in fact we could even include this in the public API.

app/services/api/v3/dashboards/filter_commodities_from_contexts.rb
```diff
@@ -0,0 +1,50 @@
...
+        end
+
+        def call
+          {data: @query, meta: @meta}
+        end
+
+        private
+
+        def initialize_query
+          @query = Api::V3::Context.
```
agnessa: let's order by name

PR #949, opened 2019-08-14
Allow filtering by nodes with same role and different node type in dashboards filter

agnessa:
https://www.pivotaltracker.com/story/show/167809756

The problem: you want to be able to filter nodes in the dashboards filtering panel by nodes selected on a different tab of the same panel, e.g. you select a biome on first tab and then go to municipalities tab. The reason this wasn't working before is because the node ids belonging to the category of the panel are treated differently than those passed as filters from different panels, which, when combined with the node type filter, produced an empty result. For example:
/sources?node_type_id=[id of municipality]&sources_ids=[id of a biome] - this would look for nodes with node type matching municipality and id matching the specified biome

The solution:  If the parameters include both node ids from same role, e.g. sources_ids sent to sources endpoint, and also a list of node type ids were passed, any sources_ids that don't match the node type ids are used as "neighbour filters" rather than "identity filters", so in the example above it would now return nodes with node type id matching municipality which are on same flow paths as the specified biome

### DIFF
--- a/app/services/api/v3/dashboards/base_filter.rb
+++ b/app/services/api/v3/dashboards/base_filter.rb
@@ -36,12 +36,26 @@ module Api
         end
 
         def apply_filters
+          adjust_node_filters
           filter_by_countries
           filter_by_commodities
           filter_by_nodes
           filter_by_node_types
           filter_by_profile_only
           filter_by_self
+        end
+
+        def adjust_node_filters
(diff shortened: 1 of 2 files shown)


agnessa:
FYI @sorodrigo - this fixes the issue you noticed the other day to do with filtering sources by sources from different tabs in the dashboards filter

PR #956, opened 2019-08-19
Feature/flows attribute

jesusgonzalezrivera:


app/services/api/v3/flows/filter.rb
```diff
@@ -123,7 +128,7 @@ def initialize_selected_nodes
             Api::V3::Flow.
               select('true').
               joins(:flow_quants).
-              where('flow_quants.quant_id' => @resize_quant.id).
+              where('flow_quants.quant_id' => @cont_attribute.id).
```
agnessa: I think it should be @cont_attribute.original_id

agnessa:
Looks good! Ready for front end changes.

PR #960, opened 2019-08-22
Feature/top supply cards

Bluesmile82:
This PR:

- Adds the top supply cards
- Creates some links from them to the Sankey (
- Renames ToolSelector to Explore and Explore to LegacyExplore
- Improves loading of topCountryNodes in the World map so they are loaded when we go to the second step
- Extract some logic out of the WorldMap component
- Fix some styles on the explore page to make more space for the map
- Address other PR comments

Still missing:
-Some params are missing link indicators
-Data for the cards is mocked
- Fetch to the backend for the cards is not prepared (We're missing some final decisions)

image


sorodrigo:
I think when navigating to the sankey and going back, the page should be either at the initial state or should be displaying the world map flows

sorodrigo:
Indonesia - crude palm oil is not showing flows

frontend/scripts/react-components/explore/top-cards/top-cards.scss
```diff
@@ -0,0 +1,44 @@
...
+  }
+}
+
+.c-top-card {
+  height: 100px;
+  width: 285px;
+  border-radius: 5px;
+  background-color: $white;
+  display: flex;
+  flex-direction: column;
```
sorodrigo: no need to use flex here

frontend/scripts/react-components/tool/tool.thunks.js
```diff
@@ -5,8 +5,4 @@ export const loadDisclaimerTool = dispatch => dispatch(loadDisclaimer());
 
 export const resizeSankeyTool = dispatch => dispatch(resize());
 
-export const loadToolInitialData = (dispatch, getState, bag) => {
```
sorodrigo: https://www.youtube.com/watch?v=UKfceBq-j9Y

If we remove this, then every time we update the url we're refetching stuff.

Bluesmile82:
Ready again!

sorodrigo:
I just noticed that the countries aren't being highlighted in step 2.
step2

PR #964, opened 2019-08-27
Feature/top nodes

jesusgonzalezrivera:
EXAMPLE:
http://localhost:3000/api/v3/nodes_stats?contexts_ids=1&column_id=8&attributes_ids=29

`top_destinations` has been renamed to `node_stats` giving that the endpoint accepts `node_type_id` `column_id` and `attribute_ids` as parameters, so it is not only related to the destinations and it can also return information for multiple attributes.

`node_stats` can replace the current `top_nodes` endpoint. However it does not contain `include_domestic_consumption` filter, so it is necessary to add it before totally replacing it.


agnessa:
I got `could not create unique index "nodes_stats_mv_context_id_quant_id_node_id_node_type_id_idx"` when running the migration. maybe it also needs year?

jesusgonzalezrivera:
> I got `could not create unique index "nodes_stats_mv_context_id_quant_id_node_id_node_type_id_idx"` when running the migration. maybe it also needs year?

You are right, I have added year on the last commit. There is a maximum length on the name for the indexes, so I have needed to remove `_id` for the relations included on the index on the index name.

app/services/api/v3/nodes_stats/response_builder.rb
```diff
@@ -0,0 +1,94 @@
...
+        def call
+          initialize_nodes_stats
+          filter_name = @commodity_id ? 'commodity_id' : 'context_id'
+          filter_values =
+            @commodity_id.present? ? [@commodity_id] : @contexts_ids
+          filter_values.map do |filter|
+            {
+              filter_name => filter,
+              attributes: @quants_ids.map do |quant_id|
+                quant = Api::V3::Quant.find(quant_id)
```
agnessa: If the param name is called attribute_ids, these should be ids from attributes_mv, not quants; also, perhaps it would be a bit more consistent to call it attributes_ids (analogically to contexts_ids) and in both cases of attributes_ids and contexts_ids to allow them to be passed as comma-separated strings rather than arrays. I know arrays are more idiomatic in Rails, but at some point we started to switch to comma separated (e.g. in dashboards endpoints) because they're shorter and in some cases that matters.

app/services/api/v3/nodes_stats/response_builder.rb
```diff
@@ -0,0 +1,94 @@
...
+                targets = @nodes_stats.where(quant_id: quant_id)
+                {
+                  id: quant_id,
```
agnessa: attribute id here

agnessa:
@jesusgonzalezrivera have a look at my last commit, what I tried to achieve here is that frontend can pass the id from attributes_mv (in the same way as for the sankey and dashboards endpoints) and underneath we translate to quant id for querying the mview. Let me know if you think it doesn't make sense!

Bluesmile82:
Sorry for picking this late but is the context going to return the default year stats or should we add another param to request the year that we need for each context?

PR #968, opened 2019-09-03
Fix/docz v2

Bluesmile82:
Migrate from docz to storybook.
We will still have to find a solution to move the mdx documentation. We can use github pages to organize the docs and link storybook for the components. That will be done in a different PR

image


sorodrigo:
There's some examples and components which seem to be broken. Not in scope of this PR, I love the new storybook 😬

PR #969, opened 2019-09-03
Use header indicators on basic profiles info

Bluesmile82:
This PR changes the basic info rendering on profiles (actor and places) so we render the indicators coming from the headerAttributes instead of the former ones.

Some changes on the backend might be done in the future to get rid of the NUM_DECIMALS list that has hardcoded decimal numbers for the indicator values.

image


frontend/scripts/react-components/profiles/actor-summary.component.jsx
```diff
@@ -49,74 +138,14 @@ class ActorSummary extends React.PureComponent {
...
-                      className="stats-text"
-                    >
-                      NONE
-                    </Text>
-                  </div>
-                )}
-              </div>
-            )}
-          </div>
+          {!isEmpty(
```
sorodrigo: isEmpty could be replaced with a few more lines, besides that, using an expression inside isEmpty is weird.

frontend/scripts/react-components/profiles/place-summary.component.jsx
```diff
@@ -49,9 +46,6 @@ function PlaceSummary(props) {
     { name: capitalize(biomeName), label: 'Biome' },
     { name: capitalize(stateName), label: 'State' }
   ];
-  const commodityAreaValue = formatValue(commodityArea, 'area');
-  const areaValue = formatValue(area, 'area');
-  const commodityProductionValue = formatValue(commodityProduction, 'tons');
```
sorodrigo: I think that if commodityProduction is tons here, the new value in the constants should also be tons.

frontend/scripts/react-components/profiles/place-summary.component.jsx
```diff
@@ -73,55 +67,28 @@ function PlaceSummary(props) {
   );
 
   const renderStats = () =>
-    (areaValue !== '-' || commodityAreaValue !== '-' || commodityProductionValue !== '-') && (
+    !isEmpty(headerAttributes) &&
```
sorodrigo: we can replace isEmpty with a few more lines

Bluesmile82:
Ready again!

PR #991, opened 2019-09-10
Feature/quick facts

agnessa:
https://www.pivotaltracker.com/story/show/167778490

```
{
   "data":[
      {
         "countryId":12,
         "facts":[
            {
               "year":2017,
               "attributeId":29,
               "total":38181449.8828643
            }
         ]
      },
      {
         "countryId":23,
         "facts":[
            {
               "year":2017,
               "attributeId":29,
               "total":1603570.977733
            }
         ]
      },
      {
         "countryId":27,
         "facts":[
            {
               "year":2017,
               "attributeId":29,
               "total":114599167.987056
            }
         ]
      },
      {
         "countryId":176,
         "facts":[
            {
               "year":2017,
               "attributeId":29,
               "total":10477999.9998733
            }
         ]
      }
   ],
   "meta":{
      "attributes":[
         {
            "id":29,
            "displayName":"Trade volume",
            "unit":"t",
            "tooltipText":"Amount of the traded commodity (tons). The total value is a composite of traded sub-products, which are converted to their original raw equivalents and then aggregated. For example exports of soy cake and soy oil are converted to soybean equivalents in the areas of production, where farmers are just producing beans."
         }
      ]
   }
}
```

### DIFF
--- a/app/controllers/api/v3/commodities_controller.rb
+++ b/app/controllers/api/v3/commodities_controller.rb
@@ -4,11 +4,29 @@ module Api
       include Api::V3::ParamHelpers
 
       skip_before_action :load_context
+      before_action :load_commodity, only: [:countries_facts]
       before_action :set_collection, only: [:index]
 
       def index
         render json: @collection,
                each_serializer: Api::V3::Dashboards::CommoditySerializer
+      end
+
+      def countries_facts
+        facts = Api::V3::SupplyChainCountriesFacts.new(
+          @commodity.id
+        )
+
+        serialized_attributes = ActiveModelSerializers::SerializableResource.new(
+          facts.attributes,
+          each_serializer: Api::V3::AttributeSerializer,
+          root: :attributes
+        )
+
+        render json: facts.facts,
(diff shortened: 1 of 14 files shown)


Bluesmile82:
Isn't this the same endpoint as the node_stats? Why do we need an additional endpoint?

agnessa:
I don't think node stats can give us stats per country - it returns data for top 10 nodes (destinations), whereas this gives a total per each country in a commodity. @jesusgonzalezrivera ?

Bluesmile82:
Ok, I think I was confused about the source-destination. Then we won't need extra indicators in the top 10, Just volume. @jesusgonzalezrivera  

But maybe we will need some other indicators in the future in this endpoint. They can come as some other objects in the data.

jesusgonzalezrivera:
Exactly, it is like Agnieska says, `node_stats` for destination countries and `countries_facts` for source country

agnessa:
For now this returns just one fact per country (Volume); will be up to 2 facts going forward

PR #995, opened 2019-09-10
Feature/quick facts ui

Bluesmile82:
Add quick facts indicator bubbles and style. Data is mocked

image
image


sorodrigo:
I think the bubble is not centered, also the text within the bubble looks weird, maybe we should have it in two lines?
Captura de pantalla 2019-09-11 a las 10 32 34

Also BTW, are we using the right projection 😅 ?

Bluesmile82:
We are using Robinson which is more accurate than Mercator, the one on the design. It's kind of hard to keep the map width in control with the variable height. I will try

PR #997, opened 2019-09-10
Feat: basemaps menu

sorodrigo:
This PR removes the old basemaps menu and adds a new one. It also removes the basemap restrictions. Now users can use any desired basemap even if they have the choropleth active.
basemaps


frontend/scripts/react-components/tool/basemaps/basemaps.scss
```diff
@@ -25,10 +25,16 @@
...
+    &:hover, &:active {
+      background-color: $light-gray;
+    }
+
     > .icon {
       fill: $charcoal-grey;
       height: 16px;
       width: 16px;
+      margin: 0;
+      transform: translateY(1px);
```
Bluesmile82: why using translate and not a margin-top?

sorodrigo: Good question. I think using a transform is cleaner. It’s only applied on the item. A margin shifts things around. Also it’s the true purpose of the property.

PR #1010, opened 2019-09-16
Feat: react tool map legend

sorodrigo:
This PR adds the new legend built entirely with react.


frontend/scripts/react-components/tool/legend/choropleth-legend.component.jsx
```diff
@@ -0,0 +1,99 @@
...
+            {[...bucket[0]].reverse().map(value => (
+              <li>{abbreviateNumber(value, 3)}</li>
+            ))}
+            {[...bucket[1]].reverse().map(value => (
+              <li>{abbreviateNumber(value, 3)}</li>
+            ))}
+          </ul>
+        )}
+
+        {!isBivariate && (
```
Bluesmile82: Ternary maybe

frontend/scripts/react-components/tool/legend/choropleth-legend.component.jsx
```diff
@@ -0,0 +1,99 @@
...
+        )}
+        {isBivariate && (
+          <svg className="icon icon-bidimensional-legend-arrows">
+            <use xlinkHref="#icon-bidimensional-legend-arrows" />
+          </svg>
+        )}
+
+        <ul className="bullets">
+          <li>
+            <div className="bullet color-dfdfdf" style={{ background: '#DFDFDF' }} />
```
Bluesmile82: Why do we need the class and the style?

frontend/scripts/react-components/tool/legend/choropleth-legend.component.jsx
sorodrigo: I dont love jsx ternaries, I have more difficulty reading them...

frontend/scripts/react-components/tool/legend/choropleth-legend.component.jsx
sorodrigo: good question, have no answer to it. I just translated some vanilla code. Will give it a go at cleaning this, I just didnt wanted to spend much time on this.

sorodrigo:
I've added the missing map warnings and cleaned some of the styles. Ready to be merged

Bluesmile82:
There is a bug here. If we change one column type the layers disappear and you are unable to select more

sorodrigo:
You are able to select contextual layers, the choropleth layers are being set, but aren't being shown because they're only available for municipalities. The fix is to disable and remove selected layers from the state, when not in municipalities column. I managed to replicate this bug in production, and makes more sense to fix it when we add the layers modal

Bluesmile82:
Ok, let's wait to the modal then

PR #1012, opened 2019-09-17
Feature/nodes stats endpoint

Bluesmile82:
- Use the new nodes_stats endpoint to retrieve the flows for the world maps
- The keys for the top nodes have been removed as the browser will cache the responses
- The top nodes keys are still used on the legacy-explore. That has a different endpoint, action, and reducer. And should be removed when is not used anymore

### DIFF
--- a/frontend/scripts/react-components/tool-links/tool-links.fetch.saga.js
+++ b/frontend/scripts/react-components/tool-links/tool-links.fetch.saga.js
@@ -80,7 +80,9 @@ export function* getToolLinksData() {
     }
   } finally {
     if (yield cancelled()) {
-      if (NODE_ENV_DEV) console.error('Cancelled');
+      if (NODE_ENV_DEV) {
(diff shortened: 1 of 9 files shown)


frontend/scripts/react-components/tool-links/tool-links.fetch.saga.js
```diff
@@ -196,9 +186,7 @@ export function* getMissingLockedNodes() {
   const nodesIds = Array.from(lockedNodes).filter(lockedNode => !nodes[lockedNode]);
 
   if (nodesIds.length === 0) {
-    if (NODE_ENV_DEV) {
-      console.log('No missing nodes.');
-    }
+    if (NODE_ENV_DEV) console.log('No missing nodes.');
```
sorodrigo: http://giphygifs.s3.amazonaws.com/media/1M9fmo1WAFVK0/giphy.gif

PR #1017, opened 2019-09-18
We need the vanilla tooltip as we use an ejs

Bluesmile82:
The tooltip was missing on the layers info. As we use an ejs template in the map-dimensions component we cant use the react tooltip so we need to load the vanilla tooltip to the tool.

image


sorodrigo:
My bad. Thanks for taking care of this.

PR #1018, opened 2019-09-18
Feature/mobile top cards

Bluesmile82:
- Add mobile dropdowns
- Make top cards scrollable
- Fix responsive. The height doesn't change anymore but you are able to scroll in very short screens

image
responsive


sorodrigo:
See the attached pictures, also i can't scroll on the cards below.
Screen Shot 2019-09-20 at 17 33 38
Screen Shot 2019-09-20 at 17 34 43

sorodrigo:
nevermind the scroll, its a merge issue.

sorodrigo:
Also to take into account. 
IMG_0135
IMG_0134

Bluesmile82:
Merged develop (Rebasing was too much). There is still an edge case when we have very short screens and it turns into the mobile selectors. But I think it's ok for now. Please try it and let's merge

PR #1033, opened 2019-09-23
Feature/tool toggle

Bluesmile82:
- Add LayoutArrows component
- Move SET_SANKEY_SIZE to toolLayers reducer
- Fix tests

image
image
image
change layout


frontend/scripts/router/nav-links.js
```diff
@@ -12,14 +14,14 @@ if (ENABLE_REDESIGN_PAGES) {
       name: 'Supply Chain',
       page: {
         type: 'tool',
-        payload: { serializerParams: { isMapVisible: false } }
+        payload: { serializerParams: { toolLayout: TOOL_LAYOUT.right } }
```
sorodrigo: this probably should be splitted

frontend/scripts/tests/tool-layers/reducer.test.js
```diff
@@ -133,19 +134,11 @@ test(SELECT_CONTEXTUAL_LAYERS, () => {
   expect(newState).toMatchSnapshot();
 });
 
-describe(TOGGLE_MAP, () => {
-  it('Toggles the map visibility if forceState is null', () => {
+describe(CHANGE_LAYOUT, () => {
```
sorodrigo: there's a new action that sets sankey size, could we add a test?

sorodrigo:
This is great! I have some suggestions:

1. Create a `<SplittedView />` component that renders children in two slots (left and right) to prepare app for the dashboards.
2. Disable layers on map, when map is collapsed for improved performance. (Maybe the same with links for sankey?)
3. Add a transparent div on top of the collapsed section so that the users can click on the visible area to go back to the splitted view.
4. If I go into the tool and change to map view, then go to explore and click a link. The link will open in the map view (previously set state). I think cards should either contain the default layout view as a param or always go to the default view (splitted view).

Bluesmile82:
About 4. The cards could help you see the Sankey view or the data view. We don't have that switch yet. Probably entering the map in the split view always would be the best option but we can do that when we fix the top cards link.

sorodrigo:
Ok! but is that case the backend should remain agnostic to the layout.

Bluesmile82:
Rebased! Please check the changes to see if I missed something from develop

PR #1043, opened 2019-09-26
Feat: timeline component

sorodrigo:
New component to select years.


sorodrigo:
update range/year toggle options order, make sure it works with feature flags.

PR #1051, opened 2019-10-01
Feature/indicator selection modal

Bluesmile82:
- Create new indicator modal
- Create selector with the same look as the former dropdown

Note: the z-index and boundaries have been changed as they only affect this modal and couldn't be positioned in another way

image
image


Bluesmile82:
Changes applied! About the insideTooltip. We were only using it on this particular tooltip on this particular case so I can't find a reason to restore it

PR #1060, opened 2019-10-03
Feature/quick sankey cards

jesusgonzalezrivera:
Sankey card links entity has the `level` field to represent the three levels available through the design. There is a maximum number of sankey card links for each level, so:
- `level = 1` - This level is used when the user has not selected commodity and country. It permits a maximum of 4 Sankey card links.
- `level = 2` - When the user has selected the commodity. It permits a maximum of 4 Sankey card links for each commodity.
- `level = 3` - When the user has selected commodity and country (context). It permits a maximum of 4 Sankey card links for each context.

### DIFF
--- a/app/admin/api/v3/sankey_card_link.rb
+++ b/app/admin/api/v3/sankey_card_link.rb
@@ -15,11 +15,12 @@ ActiveAdmin.register Api::V3::SankeyCardLink, as: 'SankeyCardLinks' do
     f.semantic_errors
     inputs do
       input :link_param, input_html: {value: f.object.link}, as: :string, required: true
-      input :title, as: :string, required: true
-      input :subtitle, as: :string
-      input :level1, as: :boolean
-      input :level2, as: :boolean
-      input :level3, as: :boolean
+      input :title, as: :string, required: true,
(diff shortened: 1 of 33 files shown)


app/admin/api/v3/sankey_card_link.rb
```diff
@@ -0,0 +1,53 @@
...
+  end
+
+  index do
+    column('Title', sortable: true, &:title)
+    column :subtitle
+    column('Link', sortable: true) do |sankey_card_link|
+      link_to(sankey_card_link.link, sankey_card_link.link)
+    end
+    column :level do |sankey_card_link|
+      [1, 2, 3].select { |n| sankey_card_link.send("level#{n}") }.join(', ')
```
agnessa: I think it would be better to avoid "magic numbers" in favour of a constant, e.g. `Api::V3::SankeyCardLink::LEVELS = [1,2,3].freeze` and use that whenever we need to enumerate the levels

jesusgonzalezrivera:
You are completely right. I have fixed the code with the comments you have written and I have also added the logic for updating `query_params` when any of the relations changes.

PR #1064, opened 2019-10-04
Feat: recolor by tooltip legend

sorodrigo:


frontend/scripts/react-components/tool/sankey/recolor-by-legend/recolor-by-legend.scss
```diff
@@ -0,0 +1,87 @@
...
+        display: none;
+        margin-right: 10px;
+
+        &::before {
+          display: inline-block;
+          width: 11px;
+          height: 11px;
+          margin-right: 4px;
+          border: 2px solid $white;
+          border-radius: 99px;
```
Bluesmile82: Why 99px instead of 50%? And why this white border?

Bluesmile82:
This is more a design decision. Shouldn't the 24 be 24%. I think it would be clearer like this

image

PR #1087, opened 2019-10-14
Feature/public api attributes

jesusgonzalezrivera:


app/services/api/public/attributes/response_builder.rb
```diff
@@ -0,0 +1,83 @@
...
+            joins('INNER JOIN countries ON ' \
+                  'countries.id = contexts.country_id').
+            joins('INNER JOIN commodities ON ' \
+                  'commodities.id = contexts.commodity_id').
+            group(*flow_attributes_group_clause)
+        end
+
+        def flow_attributes_select_clause
+          [
+            'name',
```
agnessa: let's also return the attribute id

PR #1089, opened 2019-10-14
Fix/firefox layer menu

Bluesmile82:
- Fix firefox layer menu. The height was reduced as there was a difference in the calculation between browsers.
- Flip node menu if it crosses the node bounds

image
image

### DIFF
--- a/frontend/scripts/react-components/tool/sankey/node-menu.component.jsx
+++ b/frontend/scripts/react-components/tool/sankey/node-menu.component.jsx
@@ -1,12 +1,21 @@
-import React from 'react';
+import React, { useMemo } from 'react';
 import PropTypes from 'prop-types';
 import cx from 'classnames';
 import Text from 'react-components/shared/text';
 
 import './node-menu.scss';
 
+const ITEM_HEIGHT = 33;
+
 function NodeMenu(props) {
-  const { options, menuPos, isVisible } = props;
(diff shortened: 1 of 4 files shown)


frontend/scripts/react-components/tool/sankey/node-menu.component.jsx
```diff
@@ -1,12 +1,21 @@
...
 
 import './node-menu.scss';
 
+const ITEM_HEIGHT = 33;
+
 function NodeMenu(props) {
-  const { options, menuPos, isVisible } = props;
+  const { options, menuPos, isVisible, containerRef } = props;
+  const menuHeight = useMemo(() => ITEM_HEIGHT * options?.length, [options]);
+  const flipped = useMemo(() => {
```
sorodrigo: I would drop memo, because everytime the calculations need to be done the memo function will be called. There's no advantage here.

frontend/scripts/react-components/tool/sankey/node-menu.scss
```diff
@@ -24,7 +24,12 @@
       box-shadow: unset;
 
       .options {
-        display: block;
+        display: flex;
+        flex-direction: column;
+
+        &.flipped {
+          flex-direction: column-reverse;
```
sorodrigo: elegant asf

PR #1093, opened 2019-10-15
Add extra columns to Sankey

Bluesmile82:
- You can access a new option 'See municipalities' on the biome node menu
- You can close the municipalities column by selecting the 'Close municipalities' option on the biome node
- You can change the biome column (in this case) and the extra column will only appear if possible
- You can interact with the new column and the choropleth will react accordingly
- The biome filter dropdown is removed
- Legacy is still supported

Extra: 
- Serialized props are now in the serializers files (They were duplicated and hard to access)
- Changed name to the selectedBiomeFilter => extraColumnNodeId if the ENABLE_REDESIGN_PAGES is active

Note:

- There is a task to run when this PR is merged to update the URLs on the backend

image


frontend/scripts/react-components/tool-links/tool-links.selectors.js
```diff
@@ -49,42 +61,58 @@ export const getSelectedRecolorBy = createSelector(
...
       return null;
     }
     const visibleNodes = getVisibleNodesUtil(links, nodes, selectedColumnsIds);
-    const visibleColumns = new Set(visibleNodes.map(node => node.columnId));
-    return visibleColumns.size === NUM_COLUMNS ? visibleNodes : null;
+    const visibleNodesWithCorrectedColumn = visibleNodes.map(node => {
+      if (!toolColumns || !extraColumnId || toolColumns[node.columnId].filterBy !== extraColumnId) {
+        return node;
+      }
+      return { ...node, columnId: extraColumnId };
```
sorodrigo: arent you overriding the columnId here?

sorodrigo: is there a way to do this without using `toolColumns`?

frontend/scripts/react-components/tool/column-selector/column-selector.component.jsx
```diff
@@ -37,18 +39,17 @@ function ColumnSelector(props) {
...
-      style={{ left: group * sankeyColumnsWidth + group * gapBetweenColumns - 8 }}
+      style={{ left: position * sankeyColumnsWidth + position * gapBetweenColumns - 8 }}
     >
       {content}
     </div>
   );
 }
 
 ColumnSelector.propTypes = {
-  group: PropTypes.number.isRequired,
```
sorodrigo: group is the name in the column object do you prefer having a local variable name instead?

frontend/scripts/react-components/tool/map/map.component.js
```diff
@@ -477,7 +476,15 @@ export default class MapComponent {
       const color = this.darkBasemap
         ? CHOROPLETH_COLORS.bright_stroke
         : CHOROPLETH_COLORS.dark_stroke;
-
+      let isFilteredOut = false;
+      if (!ENABLE_REDESIGN_PAGES) {
```
sorodrigo: I think we should filter out the polygons

frontend/scripts/react-components/tool/sankey/sankey.component.jsx
```diff
@@ -61,23 +62,59 @@ function useMenuOptions(props, hoveredSelectedNode) {
       });
     }
 
-    if (!isReExpand && hasExpandedNodesIds) {
-      return items.filter(item => item.id !== 'expand');
+    const activeColumn = toolColumns && Object.values(toolColumns).find(c => c.name === nodeType);
```
sorodrigo: Why not pass the columnId from the node?

frontend/scripts/react-components/tool/sankey/sankey.component.jsx
```diff
@@ -61,23 +62,59 @@ function useMenuOptions(props, hoveredSelectedNode) {
...
+    const selectedNode =
```
sorodrigo: same way we got the nodeType, we could get the nodeId

frontend/scripts/react-components/tool/sankey/sankey.selectors.js
```diff
@@ -200,6 +212,12 @@ export const getSankeyLinks = createSelector(
       return null;
     }
 
+    // Fix a race condition in which the links were still loading
```
sorodrigo: i think this race condition will be gone if we manage to detach the links loading from the columns loading.

frontend/scripts/react-components/tool/tool.selectors.js
```diff
@@ -1,37 +1,61 @@
...
+
+    const selectedColumns = selectedContext.defaultColumns.reduce((acc, column) => {
       let id = column.id;
       if (selectedColumnsIds && selectedColumnsIds[column.group]) {
         id = selectedColumnsIds[column.group];
       }
-      acc[column.group] = id;
+      acc.push(id);
+
+      if (columns && columns[id].filterTo && columns[id].filterTo === extraColumnId) {
```
sorodrigo: this is the reason why we cant load links and columns in parallel, also the reason why we have race conditions and why we need to include the toolColumns everywhere. What if we also persist the parent column?

frontend/scripts/react-components/tool/tool.selectors.js
```diff
@@ -1,37 +1,61 @@
...
+    return selectedColumns;
   }
 );
 
+export const getHasExtraColumn = createSelector(
+  [getExtraColumnId, getToolColumns, getSelectedColumnsIds],
+  (extraColumnId, columns, selectedColumnsIds) => {
+    if (!columns || !extraColumnId) return false;
+    const selectedColumns = Object.values(columns).filter(c => selectedColumnsIds.includes(c.id));
+    return selectedColumns.some(c => c.filterTo === extraColumnId);
```
sorodrigo: if we persist the parent column this is fixed.

frontend/scripts/reducers/helpers/getChoropleth.js
```diff
@@ -41,10 +41,15 @@ export default function(selectedMapDimensionsUids, nodes, attributes, columns, m
         bucket[0].length + 1
       );
 
+  const choroplethColumns = Object.values(columns).filter(
+    c => c.isGeo && Object.keys(nodes).some(nodeId => nodes[nodeId].columnId === c.id)
```
sorodrigo: why are we getting the columns which have nodeIds?

frontend/scripts/reducers/helpers/getChoropleth.js
```diff
@@ -41,10 +41,15 @@ export default function(selectedMapDimensionsUids, nodes, attributes, columns, m
...
+  );
+  const lastChoroplethColumn = choroplethColumns[choroplethColumns.length - 1];
   const geoNodesIds = Object.keys(nodes).filter(nodeId => {
     const node = nodes[nodeId];
-    const column = columns[node.columnId];
-    return node.geoId !== undefined && node.geoId !== null && column.isGeo;
+    return (
+      node.geoId !== undefined && node.geoId !== null && node.columnId === lastChoroplethColumn.id
```
sorodrigo: probably this is more efficient:
```js
const choroplethColumns = Object.values(columns).filter(
    c => c.isGeo
);
const lastChoroplethColumn = choroplethColumns[choroplethColumns.length - 1];
return (
  node.geoId !== undefined && node.geoId !== null && node.columnId === lastChoroplethColumn.id
);
```

frontend/scripts/react-components/tool-layers/tool-layers.selectors.js
```diff
@@ -44,11 +44,14 @@ export const getHighlightedNodesGeoIds = createSelector(
...
-    Object.values(columns).find(column =>
-      selectedColumnsIds.some(id => id === column.id && column.isGeo)
-    )
+  (columns, selectedColumnsIds) => {
+    if (!columns) return null;
+    const selectedGeoColumns = Object.values(columns).filter(
+      column => column.isGeo && selectedColumnsIds.includes(column.id)
+    );
+    // Get extraColumn if exists as it is the last geo column
+    return selectedGeoColumns[selectedGeoColumns.length - 1];
```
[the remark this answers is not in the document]
Bluesmile82: In that case, Municipalities is the last geo column (isGeo) so that will be the one selected

frontend/scripts/react-components/tool-links/tool-links.reducer.js
```diff
@@ -215,6 +213,53 @@ const toolLinksReducer = {
...
+      const notWereInColumn = id => {
+        const node = draft.data.nodes[id];
+        const column = draft.data.columns[node.columnId];
+        return column.id !== state.extraColumnId;
+      };
+      if (columnId) {
+        // Open extra column
+        if (columnNode) {
+          const parentColumn = Object.values(draft.data.columns).find(c => c.filterBy === columnId);
+          draft.expandedNodesIds = state.expandedNodesIds
```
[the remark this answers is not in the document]
Bluesmile82: They should remain the same in this case

frontend/scripts/react-components/tool-links/tool-links.reducer.js
```diff
@@ -215,6 +213,53 @@ const toolLinksReducer = {
...
+            .concat(columnNode?.id);
+        }
+      } else {
+        // Close extra column
+        draft.expandedNodesIds = state.expandedNodesIds
+          .filter(id => id !== columnNode?.id)
+          .filter(notWereInColumn);
+        draft.selectedNodesIds = state.selectedNodesIds.filter(notWereInColumn);
+      }
+      draft.extraColumnNodeId = action.payload.nodeId || null;
```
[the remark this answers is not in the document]
Bluesmile82: When we close the column the value is null. Is just to avoid assigning undefined

frontend/scripts/react-components/tool/column-selector/column-selector.component.jsx
Bluesmile82: I don't get this 🤷‍♂

frontend/scripts/reducers/helpers/getChoropleth.js
Bluesmile82: 🤷‍♂  It looks the same. We need to filter the active columns

sorodrigo: My bad you’re right

sorodrigo:
Screen Shot 2019-10-25 at 16 00 32

PR #1106, opened 2019-10-21
Feature/dashboard toggle

Bluesmile82:
I'm basing this on `feature/dashboards-in-sankey` which will be an umbrella PR for this feature

image

- Add sankey/ dashboard toggle
- Acts as a link so we have a :section  parameter in the route

Extra: 

- Bring back missing resize by dropdown


sorodrigo:
map is dissappearing in the data-view page 🤔

Bluesmile82:
Ok, it was a problem in the polygons fetch

sorodrigo:
grrrreat 🐯

PR #1113, opened 2019-10-22
Feature/duplicate dashboard

Bluesmile82:
This PR just duplicates the dashboard element removing some parts that we woulnd't need in the new integrated dashboard. Also it renames the old dashboard-element to dashboard-element-legacy


Bluesmile82:
This will be changed in the next PR so I better just include that commit in the next one

PR #1118, opened 2019-10-24
Add capistrano-nvm gem and staging config for support with server-sid…

tiagojsag:
…e nvm and userland node versions


agnessa:
@sorodrigo: you recently upgraded node on staging, is the version 12.12 going to be fine?

sorodrigo:
yup 12.12 is perfect

PR #1123, opened 2019-10-25
Removing dashboardsdemo from configuration and documentation

agnessa:
https://www.pivotaltracker.com/story/show/169148393


config/load_tests.yml
```diff
@@ -1,15 +1,12 @@
 development:
   enabled: true
-  host: http://localhost:3000
-dashboardsdemo:
-  enabled: true
-  host: https://dashboardsdemo.trase.earth
+  host: http://localhost:3000s
```
jesusgonzalezrivera: Is it correct the 's' at the end of port 3000?

PR #1146, opened 2019-11-06
Feature/integrate node panel

Bluesmile82:
Start integration of tool panel in the tool page:
- Add tools panel
- Get the context from the nodes panel and fix serialization

please check the dashboards too as we are sharing the functionality

- Known bug 🐞: When we enter the panel with a context and we select a country that doesn't provide  the commodity of the selected context the data of the commodity will be erased in the next step so we will lose the information for the selected context and there will be errors. We tried to fix this only deleting the byId but we need some other solution as we will delete the data on the change 😓 
- Another: Sometimes when we enter the panel the steps are all blank.


sorodrigo:
I think I fixed the first issue, you were using the app's context to fetch the missing items. I rollback that change and am using instead the nodes panel context, that way it's independent from the one selected in the app.

sorodrigo:
The homepage explore map is broken now, cant select a new context.

PR #1150, opened 2019-11-07
Feature/public api node data

jesusgonzalezrivera:


app/services/api/public/node/filter.rb
```diff
@@ -0,0 +1,125 @@
...
+        def select_node_clause(node_type)
+          'ARRAY(' \
+            'SELECT JSONB_BUILD_OBJECT(' \
+              "'name', #{node_type}.name, " \
+              '\'year\', year, ' \
+              '\'value\', value) ' \
+            "FROM node_#{node_type} " \
+            "INNER JOIN #{node_type} ON #{node_type}.id = node_#{node_type}.#{node_type.singularize}_id " \
+            "WHERE node_#{node_type}.node_id = flow_nodes_mv.node_id AND " \
+              "node_#{node_type}.year = flow_nodes_mv.year" \
```
agnessa: we'll need `year = x or year is null` here to get the non-temporal values

app/services/api/public/node/filter.rb
```diff
@@ -0,0 +1,125 @@
...
+            ') AS availability',
+            'JSON_AGG(' \
+              'DISTINCT JSONB_BUILD_OBJECT(' \
+                '\'country\', contexts_mv.iso2, ' \
+                '\'commodity\', contexts_mv.commodity_name, ' \
+                "'node_quants', #{select_node_clause('quants')}, " \
+                "'node_quals', #{select_node_clause('quals')}, " \
+                "'node_inds', #{select_node_clause('inds')}" \
+              ')' \
+            ') AS node_indicators',
```
agnessa: let's not use the word "indicators" anywhere in the responses, because we are already using attributes in other endpoints and this will get too confusing I think, we'll have people asking "what's the difference between an indicator and an attribute" 😅

app/services/api/public/node/filter.rb
```diff
@@ -0,0 +1,125 @@
...
+            'JSON_AGG(' \
+              'DISTINCT JSONB_BUILD_OBJECT(' \
+                '\'country\', contexts_mv.iso2, ' \
+                '\'commodity\', contexts_mv.commodity_name, ' \
+                '\'flow_id\', flow_nodes_mv.flow_id, ' \
```
agnessa: we probably don't need the flow_id? I think it's not necessary to include that in the response, let's not expose any internal identifiers unless strictly necessary. for example, we expose the node ids in the nodes endpoints, because that allows the user to use those ids as filters in other endpoints. I think we don't currently have anywhere where they could filter by flow, and probably we won't have.

app/services/api/public/node/filter.rb
```diff
@@ -0,0 +1,125 @@
...
+                "'flow_quants', #{select_flow_clause('quants')}, " \
```
agnessa: I'd get rid of the `flow_quants`, `flow_quals` and `flow_inds` keys (and the same for `node_quants` etc), because public API users will be confused about what a qual, quant and ind is. Currently in the public attributes endpoint  `/api/public/attributes` we're not exposing this internal type, because probably there's no advantage to external users to know this. So in interest of avoiding having to respond to questions "what is an ind", I'd just mix all the values together. Currently we have the structure:
```
flowIndicators: [
  {
    country:
    flowId:
    commodity:
    flowInds: [{name: ..., year: ..., value: ...}, ...]
    flowQuals: ...
    flowQuants:
  }
]
```

maybe we could simplify:

```
flowAttributes: [
{
  country:
  commodity:
  values: [{name: quant name, value: ..., year: ...}, {name: ind name, value: ..., year: ...}]
}
]
```

Also, I think since we expose attributes ids in the `/api/public/attributes` endpoint, I guess instead of attribute names here we could use the attributes ids, which would make the response shorter. Not sure if it would make the query easier, so this is just a suggestion.

app/services/api/public/node/filter.rb
jesusgonzalezrivera: Sure! I forgot to include it

app/services/api/public/node/filter.rb
jesusgonzalezrivera: I have changed it to attributes on nodes and flows, `node_attributes` and `flow_attributes`

app/services/api/public/node/filter.rb
jesusgonzalezrivera: I have included all nodes on the same `values` property

PR #1159, opened 2019-11-13
use the same panel component in dashboards and tool

sorodrigo:
2 main things:
- reuse same panel component is both dashboards and tool
- remove need for edit mode.

smaller fixes:
- Step tracker's UI is not broken in edit mode.
- App doesnt crash on SET_CONTEXT dispatch (temporary fix, a better fix will come with columns selector)
- Fixed broken serialisation in dashboard
- Remove unnecessary duplicated code in dynamic sentence selector
- Fixed weird sentence when select all


Bluesmile82:
I found this case in which the panel is broken:

image

1. With a context selected (Brazil-Soy) open the panel
2. Select a different country (Colombia if you were in Brazil-Soy) and press continue
3. Open the panel again: Now you have Brazil selected but no commodity and you get a lot of unexpected states like missing nodes, being unable to select a step and have to go back, ...

Bluesmile82:
The close button is not closing the dashboards panel anymore if you are selecting a dashboard for the first time. It doesn't do anything on the welcome panel and it goes back to welcome in the rest.

Also, if you are editing and you change the country you have the same problem. I don't know if this is on purpose

Bluesmile82:
I'm also thinking that selecting all and not selecting any doesn't change the dynamic sentence as it doesn't change the results. But could be a bit strange for the user

sorodrigo:
Ok nice catch, done 🏁

PR #1230, opened 2020-01-08
Feat: unified nav bar

sorodrigo:
- Updates styles of the nav bar.
- Updates styles of explore page to match visuals.


frontend/styles/_settings.scss
```diff
@@ -80,6 +80,8 @@ $light-gray: #f6f6f6;
 $mid-light-gray: #e5e5e5;
 $medium-gray: #dddddd;
 $gray-text: #515253;
+$yellow-gray: #d9d9d9;
+$yet-another-gray: #999999;
```
Bluesmile82: 😆😆😆 
Maybe something more descriptive? Although we have a lot it's true. Is the only one we will remember

Bluesmile82:
Could you change the text in the logo for the Svg that we are supposed to have? The font looks different:

Currently:


Design:

Bluesmile82:
Could we reduce a bit the padding of the text on the cards so the texts shows a bit higher:

Current:
image

Design:
image

frontend/styles/_settings.scss
sorodrigo: I'm already at the point where I need to search which gray in the design equals which variable. I'm not sure it's worth it

sorodrigo:
> Could we reduce a bit the padding of the text on the cards so the texts shows a bit higher
You're comparing mobile vs desktop, this is how it actually looks like in desktop:
Screen Shot 2020-01-09 at 08 53 21

I did realise that the separator between country and commodity is larger and the font of the subtitle is bold, will make those changes

PR #1234, opened 2020-01-13
Feat/filters bar

sorodrigo:
## Description
This PR includes the new toolbar that contains the filters for the sankey page. Changes way to display tooltips and updates necessary styles.

## Testing instructions

- Test contexts with recolor by, test tooltips, test clicking each filter.
- Test contexts without recolor by, test tooltips, test clicking each filter.
- Test explore new bubbles when hovering on pills and map.


Bluesmile82:
These are not part of the PR but they will be good fo fix: 

When selecting two cards (third step) and going back with the button there is a crash:

image

On the panel editing this tab is cut:

image

Bluesmile82:
Also, we have this problem when we resize the window
image

PR #1245, opened 2020-01-21
ed25519 support

agnessa:
## Related GH issues

https://github.com/Vizzuality/trase/pull/656
https://github.com/Vizzuality/trase/pull/1032
https://github.com/Vizzuality/trase/pull/1102

## Description

[A while back we added 3 gems to the Gemfile](https://github.com/Vizzuality/trase/pull/656): `rbnacl`, `rbnacl-libsodium` and `bcrypt_pbkdf`. That helped solve an issue while deploying with Capistrano. The issue was caused by the fact that in order to use some kinds of keys those dependencies were required by net-ssh.

One of those gems, `rbnacl-libsodium`, is now discontinued, so it makes sense to revisit.

I think the root cause might have been to enable support for ed25519 keys, and if that is the case we should use the solution described in [net-ssh readme](https://github.com/net-ssh/net-ssh) and in [this issue](https://github.com/net-ssh/net-ssh/issues/478), which is to include `ed25519` and `bcrypt_pbkdf` dependencies (they are [listed as development dependencies](https://github.com/net-ssh/net-ssh/blob/master/net-ssh.gemspec), not runtime, therefore they don't get pulled in by bundler automatically).

## Testing instructions

Just try `cap staging deploy`

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -80,7 +80,7 @@ group :development do
   gem 'spring'
   gem 'spring-watcher-listen', '~> 2.0.0'
   gem 'annotate'
-  gem 'bcrypt_pbkdf', '>= 1.0', '< 2.0'
+  gem 'bcrypt_pbkdf', '~> 1.0'
   gem 'capistrano', '~> 3.11', require: false
   gem 'capistrano-bundler'
   gem 'capistrano-yarn'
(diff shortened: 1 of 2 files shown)


sorodrigo:
successfully deployed to sandbox ✅

PR #1253, opened 2020-01-27
Feat: code split reducers and hidden components

sorodrigo:
## Pivotal Tracker

Please include the link(s)
https://www.pivotaltracker.com/story/show/170874068

## Description

To improve load time we're code splitting the reducers. To do this we add a new file which will act as the actions / action creators entrypoints. This way when we import them, the reducer and the sagas will be registered and will start working.

## Testing instructions
Test all pages and features, as best as possible 😅


frontend/scripts/react-components/explore/explore.component.jsx
```diff
@@ -25,7 +25,7 @@ import {
 
 import 'react-components/explore/explore.scss';
 
-const ToolLinksModal = React.lazy(() => import('./tool-links-modal'));
+const ToolLinksModal = React.lazy(() => import(/* webpackPreload: true */ './tool-links-modal'));
```
Bluesmile82: https://caniuse.com/#feat=link-rel-preload

Are we sure we can use preload? Check firefox and now IE

Bluesmile82:
image

Bluesmile82:
This also happens with setProfilesActiveItem on profiles panel

Bluesmile82:
image
I think we should fix this at some point

Bluesmile82:
When entering the data view from the global search the data is never displayed. You get stuck on loading

frontend/scripts/react-components/explore/explore.component.jsx
sorodrigo: If it's not supported it will load it when needed, so it's ok.

frontend/scripts/react-components/home/home.component.jsx
```diff
@@ -1,10 +1,8 @@
-import React, { useEffect } from 'react';
+import React, { Suspense, useEffect } from 'react';
```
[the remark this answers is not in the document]
sorodrigo: About Suspense, it's still experimental as a data fetching and suspending UI. However, for code-splitting using lazy it's already stable (if they change it, it will have to be backwards compatible).

PR #1283, opened 2020-02-12
Fix/sankey card parameters

agnessa:
## Pivotal Tracker

https://www.pivotaltracker.com/story/show/170873735

## Description

This ended up being much bigger than initially suspected. The logic for keeping the stored links in sync with current node and node type identifiers (during a data update) was not reliable, and I replaced it with something simpler. However, it is still brittle, because of how coupled it is with front end parameter syntax.

To clean up the ActiveRecord model, I extracted the managament of query params to a separate service, which recognises the following types of params:
- array, e.g. `selectedNodesIds[]=x`
- single element, e.g. `extraColumnNodeId=x`
- comma separated list, e.g. `sources=x,y`
- comma separated list with "excluded" prefix (😨), e.g. `sources=excluded_x,y`
- the weirdest of all, custom columns param format, e.g. `selectedColumnsIds=0_1-1_2`

To store the link and return the query params to front end we don't need to look inside the values, and we return them as saved. Params are parsed during the data update process, for purposes of updating stale identifiers. Change the front end params format => this breaks.

This also includes a smaller requirement change of only allowing 3 cards for level 3.

## Testing instructions

Create a link, check clicking on the card opens the expected workspace.

* Also, running the data update process is an important check for this, I did that.


Bluesmile82:
I was creating a new link for Paraguay Soy and it shows on the CMS as Brazil Soy. Am I missing something?

image

http://localhost:8081/flows?expandedNodesIds%5B%5D=13208&expandedNodesIds%5B%5D=33&expandedNodesIds%5B%5D=329&expandedNodesIds%5B%5D=57&mapView=-25.37%2C-54.95%2C8&selectedContextId=2&selectedNodesIds%5B%5D=13208&selectedNodesIds%5B%5D=33&selectedNodesIds%5B%5D=329&selectedNodesIds%5B%5D=57&toolLayout=1

agnessa:
According to the [description in the task](https://www.pivotaltracker.com/story/show/170873735) the `selectedContextId` is no longer used; when the link is parsed, we're looking for `countries` and `commodities` parameters, and in absence of those we default to Brazil soy. Is that not right, is `selectedContextId` still used?

sorodrigo:
http://localhost:8081/flows?expandedNodesIds%5B%5D=13208&expandedNodesIds%5B%5D=33&expandedNodesIds%5B%5D=329&expandedNodesIds%5B%5D=57&mapView=-25.37%2C-54.95%2C8&selectedContextId=2&selectedNodesIds%5B%5D=13208&selectedNodesIds%5B%5D=33&selectedNodesIds%5B%5D=329&selectedNodesIds%5B%5D=57&toolLayout=1

this is a legacy url, enable redesign.

agnessa:
who are you?

PR #1331, opened 2020-03-12
Feature/opening country profiles

agnessa:
## Pivotal Tracker

https://www.pivotaltracker.com/story/show/171571075

## Description

Changes to allow opening country profiles from the profile home page, either through the search tool or the browse through.

### Database & CMS changes

No big changes were needed in the database, just relaxing a constraint on creating profiles, with the ability to specify a new profile type, 'country'. 

To enable a country profile you follow a similar procedure as for actor profile, i.e.
1. go to Profiles -> Profiles and create two profiles:
  - COUNTRY OF PRODUCTION node type with profile type = country
  - COUNTRY node type with profile type = country

    The two node types correspond to the export / import activities. Just like with actor profiles, the importer and exporter are actually separate nodes.

2. go to Profiles -> Charts and create a "basic attributes" chart for each of the two profiles:
  - parent=nil
  - identifier=country_basic_attributes
  - title=Basic attributes
  - position=0
  
  This won't do anything for now, as it's just a placeholder implementation to unblock work on the "browse through".

### Opening a country profile from the search tool

Both types of country nodes will come up separately in the search. There's a problem with display of the long node type name "country of production".

Screenshot 2020-03-11 at 20 26 44

### Opening a country profile from the search tool

This is more complicated. So far two endpoints have been amended to support this.

#### Changes to filter_meta endpoint

This endpoint is used in both the "dashboards" wizard and in the profile "browse through" to know which "node type tabs" to display on particular screens. Because it ends up working slightly differently for profiles and for dashboards, I decided to split the endpoint. As a result we have the old "dashboards" endpoint:
/api/v3/dashboards/filter_meta?countries_ids=12
and the new "profiles" endpoint:
/api/v3/profiles/filter_meta?countries_ids=12

_Please note: both these endpoints only accept 2 parameters, `countries_ids` and `commodities_ids`. No need to pass `profile_only` anymore, and passing `options_type` and `node_types_ids` is currently ineffective. Consequently, this endpoint only needs reloading if either country or commodity ids change._

The difference between the two endpoints is as follows:
- the dashboards version returns all node types per role EXCEPT country of production
- the profiles version only returns node types per role where a profile page can be found, INCLUDING country of production
- also, the dashboards version returns an extra property "prefix"

Visually, this is the dashboards version:
Screenshot 2020-03-11 at 20 37 25

This is the profiles version:
Screenshot 2020-03-11 at 20 37 38

That will allow to select a country of production node in the sources tab.

In order to support selecting country of destination, in the new tab  for importing countries, you'd implement it identically to how the import countries work in the dashboards filter.

#### Changes to profile_meta endpoint - url

Old form of the url:
/api/v3/contexts/37/nodes/18977/profile_metadata
Now redirects to:
/api/v3/profiles/18977/profile_meta?context_id=37

This change is so that the endpoint no longer requires to provide the context_id. If it is provided, it will be used; otherwise the endpoint will use `commodity_id` or `country_id` parameters and arrive at best match. This way it will be possible to open the import country profile from the "browse through", it will also simplify switching profiles using dropdowns.

#### Changes to profile_meta endpoint - dropdowns

There were multiple changes needed here to support the new dropdowns on the profile page, and some that were made in order to cleanup. These are the changes to existing properties:
- `id` - it is now the id of the node, used to be the id of an internal profile object for no good reason
- `context_node_type` - removed, don't think it was used
- `available_years` renamed to `years`

A bunch of new properties were added, that have to do with the dropdowns:
- `context_id` - currently selected context
- `country_id` - currently selected country (this is relevant for actor profiles only)
- `commodity_id` - currently selected commodity
- `commodities` - list of commodities available for the currently open profile. when switching, use the current node id + selected commodity to parametrise the next call to profile_meta
- `activity` - currently selected activity (exporter / importer). this applies to country & actor profiles.
- `activities` - list of activities of this node, e.g. the importer / exporter activity of an actor. Please note: changing the activity means opening a profile for a completely different node, and the id is given here. Use the selected node id + current commodity_id to parametrise the next call to profile_meta


Bluesmile82:
I added two commits for the frontend fixes:

- Change importer and exporter countries labels on search: Fixes the overflowing text on the search dropdown items
- Automatically select first country of production: We don't display anymore the country of production item (it's only one) and we select it by default

PR #1337, opened 2020-03-23
Feature/profile chart country top traders destinations

agnessa:
## Pivotal Tracker

https://www.pivotaltracker.com/story/show/171810026
https://www.pivotaltracker.com/story/show/171810006
https://www.pivotaltracker.com/story/show/171810038
https://www.pivotaltracker.com/story/show/171810035

## Description
First of all:
`bundle exec rake db:migrate db:country_profiles:populate`

Adds 2 new endpoints which will be used for the "top traders" and "top countries" endpoints on country profiles.

These endpoints **do not depend on the presence of `context_id`**. If `context_id` is provided it will be used and only return data for that context, but the intended use for country profiles is with `commodity_id` and to return data cross-context in case of importer profiles.

## Testing instructions

Please note the identifiers in these examples:
34 - id of Argentina exporter node
33 - id of Argentina importer node
56 - id of corn commodity
58 - id of Argentina - corn context

### EXPORTER COUNTRY

**top traders of corn in Argentina in 2015**
/api/v3/country_profiles/34/top_consumer_actors?commodity_id=56&year=2015
/api/v3/country_profiles/34/top_consumer_actors?context_id=58&year=2015
 (in case of exporter profile these calls would return the same data, because Argentina is a corn exporter only in one context)

**top destinations of Argentina soy in 2015**
/api/v3/country_profiles/34/top_consumer_countries?commodity_id=56&year=2015
/api/v3/country_profiles/34/top_consumer_countries?context_id=58&year=2015
 (in case of exporter profile these calls would return the same data, because Argentina is a corn exporter only in one context)

### IMPORTER COUNTRY

**top import traders of corn to Argentina in 2015**
/api/v3/country_profiles/33/top_consumer_actors?commodity_id=56&year=2015
note: in this case passing context_id changes the result

**top export countries of corn to Argentina**
/api/v3/country_profiles/33/top_consumer_countries?commodity_id=56&year=2015
note: in this case passing context_id changes the result


Bluesmile82:
This endpoint http://localhost:3000/api/v3/profiles/34/profile_meta?context_id=58 is giving back the charts but all their chart_type is null

frontend/scripts/utils/getURLFromParams.js
```diff
@@ -130,7 +131,7 @@ const API_ENDPOINTS = {
   },
   [GET_PROFILE_METADATA]: {
     api: 3,
-    endpoint: '/profiles/$node_id$/profile_meta'
+    endpoint: '/contexts/$context_id$/nodes/$node_id$/profile_metadata'
```
agnessa: this is the old url, it will redirect to the new one. please use the new one with ?context_id=xx or ?commodity_id=xx so that we can retire the redirect once not needed anymore.

agnessa:
> This endpoint http://localhost:3000/api/v3/profiles/34/profile_meta?context_id=58 is giving back the charts but all their chart_type are null

fixed now

Bluesmile82:
I added some frontend fix for the endpoints. I think we can merge. We could fix some other things when we have the summary endpoint merged

PR #1354, opened 2020-04-01
Remaining chart endpoints

agnessa:
## Pivotal Tracker

https://www.pivotaltracker.com/story/show/171810074
https://www.pivotaltracker.com/story/show/171810062
https://www.pivotaltracker.com/story/show/171810069
https://www.pivotaltracker.com/story/show/171810084

## Description

`bundle exec rake db:migrate`
`bundle exec rake db:country_profiles:populate` to create chart configuration

The rake task will copy configuration from existing place profiles for that country where available / applicable. 

Please note: no aggregated data present in Trase DB for country nodes at this point, as well as no external data sources are integrated - data is missing, I tried to indicate what to expect per chart.

Also, I added chart urls to profile_meta endpoint. The idea is to make it easier for front end to integrate the new endpoints, possibly using a more dynamic approach like in dashboards charts.

### Exporter: indicators

/api/v3/country_profiles/{node id}/indicators?year=2017

Notes:
- response structure is identical as for place profiles, except I removed the "state ranking" column as it makes no sense.
- data is not yet available (it should be provided through the Trase DB, so nothing we can improve here for now)

### Exporter: deforestation trajectory

/api/v3/country_profiles/58/trajectory_deforestation?commodity_id=6

Notes:
- response structure is identical as for place profiles deforestation trajectory,
- only production data available at this time (line chart)

### Importer: import trajectory

/api/v3/country_profiles/57/trajectory_import?commodity_id=6

Notes:
- response structure is identical as for place profiles deforestation trajectory,
- import value not available, so plotting import volume instead (just to give you some data to work with)

### Exporter: commodity exports

/api/v3/country_profiles/58/commodity_exports?commodity_id=6

Notes:
- not much to see here in terms of data yet, as it will mostly come from ComTrade API

### Importer: commodity imports

/api/v3/country_profiles/58/commodity_exports?commodity_id=6

Notes:
- not much to see here in terms of data yet, as it will mostly come from ComTrade API


agnessa:
This contains all the remaining endpoints, some data is still missing but hopefully there's enough to get you going with front end integration.

PR #1363, opened 2020-04-09
Feature/worldbank

jesusgonzalezrivera:
## Pivotal Tracker

[https://www.pivotaltracker.com/n/projects/2380431/stories/171875185](https://www.pivotaltracker.com/n/projects/2380431/stories/171875185)

## Description

This PR includes all the logic to get the indicators from Worldbank API periodically

## Testing instructions

In order to simulate the process that imports the Worldbank indicators periodically, we can use the following command:

```
bundle exec bin/rails runner 'Api::V3::CountriesWbIndicators::ImporterService.import'
```

### DIFF
--- a/app/services/api/v3/countries_wb_indicators/api_service.rb
+++ b/app/services/api/v3/countries_wb_indicators/api_service.rb
@@ -35,22 +35,29 @@ module Api
         private_class_method def self.indicator_request(name, iso_code)
           uri = URI(
             "#{ENV['WORLD_BANK_API_URL']}/v2/country/#{iso_code}/indicator/" \
-            "#{INDICATORS[name]}"
+            "#{INDICATORS[name]}?format=json&per_page=10000"
           )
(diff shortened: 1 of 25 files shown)


app/services/api/v3/countries_wb_indicators/api_service.rb
```diff
@@ -0,0 +1,65 @@
...
+        end
+
+        def self.forested_land_area_indicators(iso_code = 'ALL')
+          indicator_request(:forested_land_area, iso_code)
+        end
+
+        private_class_method def self.indicator_request(name, iso_code)
+          uri = URI(
+            "#{ENV['WORLD_BANK_API_URL']}/v2/country/#{iso_code}/indicator/" \
+            "#{INDICATORS[name]}?format=json&per_page=10000"
```
agnessa, line 38: is it possible to filter by year range? I think WB has ~50 years of data, we need much less in Trase, so that could be a way to optimise

PR #1368, opened 2020-04-17
Feature/layer manager map

Bluesmile82:
## Description

This is the first PR for changes to move the map into Mapbox and Layer manager:
Let's merge this PRs into feature/new-map until is ready and we can merge it to develop

image

- Starts the map with layer manager and Mapbox GL
- Set map controls, attribution, legend, ...
- Add contextual layers templates (The raster templates are not working due to a permission problem. It should be fixed in S3)
- Start fitting to bounds and hovering

The contextual layers have an abstracted layer generation using the functions on the layer-utils file. These shorten the code considerably and makes them easier to edit.

## Testing instructions

Go to any map on the tool. Try the contextual layers. There is one unit layer for brazil municipalities hardcoded.


frontend/scripts/react-components/tool/mapbox-map/mapbox-map.component.jsx
```diff
@@ -0,0 +1,181 @@
...
+  }, [loaded]);
+
+  const baseLayerInfo = BASEMAPS[basemapId];
+  const baseLayer = getBaseLayer(baseLayerInfo);
+
+  const layerOrder = getLayerOrder(baseLayerInfo.id);
+  const onHover = e => {
+    const { features, center } = e;
+    if (features?.length) {
+      const geoFeature = features.find(f => f.source === 'brazil_municipalities'); // Temporary
```
tomasmoose, line 88: maybe add a `// TODO:` here so we don't forget about this in the future!

Bluesmile82: Thanks! we won't forget it as we will have to import the different unit layers and check the interaction

PR #1408, opened 2020-05-25
Feature/map interaction

Bluesmile82:
## Description

This PR:

- Adds click and hover interaction
- Fetches new pbf unit layers from the services endpoint.
- Fetches the new unit layer uploaded data from Carto DB
- Adds this unit layers data to the map tooltips. This data depends on the years selected.

- There are some regions without data and they don't show any tooltip.

Missing:
- Soy yield indicator is giving accumulated data but it differs from the value on the Sankey. We might have to change the Sankey logic.

image

## Testing instructions

- Open different map contexts (Brazil-Soy), change the unit layers (On the layers modal) or change the Sankey columns (Municipalities => States)
- The map should have a tooltip with interaction when you hover and the hovered region should be highlighted in black. 
- The selected regions on the map should be selected on the tooltip and vice versa. The selected regions on the map should be highlighted in black.


frontend/.env.sample
```diff
@@ -1,6 +1,7 @@
 PORT=3000
 NODE_ENV=development
 API_V3_URL=//trase.earth
+UNIT_LAYERS_API_URL=//trase.earth
 GOOGLE_ANALYTICS_KEY=UA-87285243-1
```
tomasmoose, line 5: is this a leak or is it fine?

Bluesmile82: It should be fine. It will be the default for production

frontend/scripts/react-components/tool-layers/tool-layers.fetch.saga.js
```diff
@@ -119,3 +120,24 @@ export function* getMapDimensions(selectedContext, selectedYears) {
     }
   }
 }
+
+export function* getUnitLayerData(params) {
+  const { selectedGeoColumnId, selectedUnitIndicatorIds, iso2 } = params;
+  const url = `//p2cs-sei.carto.com/api/v2/sql?q=SELECT node_id, geo_id, attribute_id, json_object_agg(COALESCE(year, 0 ), total ORDER BY year) as years FROM (SELECT node_id, geo_id, attribute_id, year, SUM(value) AS total FROM map_attributes_values WHERE node_type_id = ${selectedGeoColumnId} and attribute_id IN (${selectedUnitIndicatorIds.join(',')}) and iso2 = '${iso2}' GROUP BY node_id, geo_id, attribute_id, year) s GROUP BY node_id, geo_id, attribute_id ORDER BY geo_id, node_id, attribute_id`;
```
tomasmoose, line 126: i recommend using something like `oneLineTrim template literal` from 'common-tags' so you can separate these into one line for easier readability https://www.npmjs.com/package/common-tags#onelinetrim 

@Bluesmile82

frontend/scripts/react-components/tool/mapbox-map/mapbox-map.selectors.js
```diff
@@ -0,0 +1,102 @@
...
+      value += unitLayerValues.years[i];
+    }
+  }
+
+  if (!value && unitLayerValues.years['0']) { // Non-temporal indicator values are stored in year 0
+    return unitLayerValues.years['0'];
+  }
+
+  if (dimension.unit === '%') {
+    value /= (lastYear - firstYear + 1);
```
tomasmoose, line 41: fancy!

PR #1460, opened 2020-07-03
Feature/new mapbox map

Bluesmile82:
🎉 Welcome to the new Mapbox map 🎉

Don't merge yet until we have some feedback and its approved

This map replaces the leaflet map with a Mapbox / Layer Manager map. All the former features of the map should be working and also:

Includes logistics map into the tool map with all the features (Heatmap + points)
Adds logistic downloads into the downloads page
Contextual layers now are vector layers with Mapbox styles applied (hardcoded on the app)
Unit layers now have vector layers imported from a server that is in sandbox / production. This is needed as the parsed geoJson had some artifacts on different levels.
The data for the unit layers now is imported from a CSV on the backend and merged into these layers.
The map is no longer a bottleneck for the loading of the page and we don't need to wait for selecting different contexts or nodes.

Still needed:

- Delete the former map code
- Style downloads tabs
- Check change of zoom when we have a selection (Goes back to a different zoom level)
- Check Indonesia Mills info on the dropdown


sorodrigo:
yaaaaaaaaay!!!! 😍 😍 😍 😍 😍 😍 😍 😍 😍 😍 😍 😍 😍 😍 😍 😍 😍 😍 😍 😍 😍 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏

PR #1473, opened 2020-07-14
Feat/country profiles

tomasmoose:
## Description

Included some new/redefined widgets for country profiles, included a checkbox as there are multiple tasks related to this. 

- [x] Include summary widget with new country profile data
https://www.pivotaltracker.com/story/show/171568760

- [x] Map widget 
https://www.pivotaltracker.com/story/show/171568576

- [x] Table widget 
https://www.pivotaltracker.com/story/show/171568384

- [x] New line widget (based on previous widgets)
https://www.pivotaltracker.com/story/show/171568521

- [x] Data for each of the widgets
https://www.pivotaltracker.com/story/show/171568904

## Testing instructions

Test both source and importer country profiles and make sure widgets appear. And also switch context and year and profile pages should update with new data.


frontend/scripts/react-components/profile/profile-components/line/line.component.jsx
```diff
@@ -213,9 +234,34 @@ class Line extends Component {
...
+        const x0 = x.invert(mouse(this)[0]);
+        const pointIndex = bisectDate(lineValuesWithFormat, x0, 1);
+        const d0 = lineValuesWithFormat[pointIndex - 1];
+        const d1 = lineValuesWithFormat[pointIndex];
+        const dxx = x0 - d0.date > d1.date - x0 ? d1 : d0;
+
+        showTooltipCallback(dxx, d3_event.clientX + 10, d3_event.clientY + scrollOffset() + 10);
+      }
+
+      if (showTooltipCallback !== undefined && type === 'line') {
```
tomasmoose, line 253: This could be applied to the multi line charts aswell later but will create a seperate task for it https://www.pivotaltracker.com/story/show/173829925

Bluesmile82:
I think we could adjust the magnitude that we show on the axis tics here. They are a bit difficult to read. Maybe using d3 format and adjusting the unit. Or if ha is the standard at least separate the thousands with some dots or commas

image

Bluesmile82:
There is something a bit weird or maybe I'm not understanding the data. It seems like top imports are different when shrimp and coffee are selected, even if the year and country are the same. Check for example Shrimp on the Coffee selection:
image
image
It

frontend/scripts/react-components/profile/profile-components/table/table.component.jsx
```diff
@@ -198,11 +270,16 @@ class ProfilesTable extends Component {
 
     return (
       <table>
+        {type === 't_head_top_imports' && this.renderTopImportsHeader()}
+        {type === 't_head_top_imports' && this.renderTopImportsTable()}
         {type === 't_head_places' && this.renderPlacesTableHeader()}
```
Bluesmile82, line 275: I really don't like this part. I hope we can improve it sometime

tomasmoose: i agree, i think all this logic should be re-written with re-charts, so we have the same logic as the dashboards

frontend/scripts/react-components/profile/profile-components/line/line.component.jsx
```diff
@@ -118,11 +129,21 @@ class Line extends Component {
...
-      // eslint-disable-next-line default-case
+      if (highlightYear) {
+        d3Container
+          .append('g')
+          .attr('class', 'highlight')
+          .append('rect')
+          .attr('width', 50)
+          .attr('height', chartHeight + chartMargin.bottom)
+          .attr('fill', 'rgba(255,255,255,0.5)')
+          .attr('x', x(new Date(year, 0)) - 25)
```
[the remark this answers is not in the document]
tomasmoose: you are right, updated

agnessa:
> There is something a bit weird or maybe I'm not understanding the data. It seems like top imports are different when shrimp and coffee are selected, even if the year and country are the same. Check for example Shrimp on the Coffee selection:
> 
> image
> image

I think that's because something is wrong with the selected year, which has to do with switching between commodities that have different years available in Trase DB. E.g. start with Palm oil and 2015 (only available year) selected - soy has rank 49, which is the value for 2015. Now switch commodity to Palm kernel, which goes up until 2018. The year selected is still 2015, but the request for commodities fetches values for 2018 (soy rank 53), because it is malformed:
Screenshot 2020-07-28 at 09 15 43

PR #1481, opened 2020-07-22
Feat/country profile links

tomasmoose:
## Pivotal Tracker

https://www.pivotaltracker.com/story/show/171568692

## Description

Profile links for country profiles, reusing the ones we had previously only parsing is a bit different for the links. Also fixed a small issue where we try to update state on a unmounted component. `banner.component` 

## Testing instructions

Test both export/import countries, make sure links correspond to the correct page.


tomasmoose:
Will block this until we have a solution for country (importers) as we dont have context.

tomasmoose:
We will only show the links if we have a context @Bluesmile82

Bluesmile82:
Please, solve the conflict so we can merge

Bluesmile82:
Should we merge this? Just solving the conflict before

tomasmoose:
> Should we merge this? Just solving the conflict before

hi will solve this today @Bluesmile82

PR #1483, opened 2020-07-23
Fix issue where we where using the wrong data on mini sankey

tomasmoose:
## Pivotal Tracker

none

## Description

This pr fixes an issue where we where displaying wrong data on the mini sankey, we now are using the chart.url to fetch the correct information as in (countries/companies) 

**Example issue, here we should display traders not countries** 
Screenshot 2020-07-23 at 10 46 31

## Testing instructions

Check all profile types and make sure we are using the correct context in the sankey.

### DIFF
--- a/frontend/scripts/react-components/profile/profile-widgets/top-consumers-widget.component.jsx
+++ b/frontend/scripts/react-components/profile/profile-widgets/top-consumers-widget.component.jsx
@@ -3,110 +3,123 @@ import PropTypes from 'prop-types';
 import Widget from 'react-components/widgets/widget.component';
 import MiniSankey from 'react-components/profile/profile-components/mini-sankey/mini-sankey.component';
 import { withTranslation } from 'react-components/nav/locale-selector/with-translation.hoc';
-import {
-  getSummaryEndpoint,
-  GET_PLACE_TOP_CONSUMER_ACTORS,
-  GET_PLACE_TOP_CONSUMER_COUNTRIES,
-  GET_COUNTRY_TOP_CONSUMER_COUNTRIES
-} from 'utils/getURLFromParams';
+import { getSummaryEndpoint } from 'utils/getURLFromParams';
 import ShrinkingSpinner from 'react-components/shared/shrinking-spinner/shrinking-spinner.component';
 import Heading from 'react-components/shared/heading/heading.component';
 import ProfileTitle from 'react-components/profile/profile-components/profile-title.component';
 
 const TranslatedMiniSankey = withTranslation(MiniSankey);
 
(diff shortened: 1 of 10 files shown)


frontend/scripts/react-components/profile/profile-widgets/top-consumers-widget.component.jsx
```diff
@@ -3,110 +3,123 @@ import PropTypes from 'prop-types';
...
 import ShrinkingSpinner from 'react-components/shared/shrinking-spinner/shrinking-spinner.component';
 import Heading from 'react-components/shared/heading/heading.component';
 import ProfileTitle from 'react-components/profile/profile-components/profile-title.component';
 
 const TranslatedMiniSankey = withTranslation(MiniSankey);
 
-function TopConsumersWidget (props) {
-  const { onLinkClick, year, nodeId, contextId, type, testId, title, commodityName, profileType } = props;
-  const handleLinkClick = (_, { query }) => {
+class TopConsumersWidget extends React.PureComponent {
```
tomasmoose, line 13: changed to class based component so we are consistent with the other widgets, this might change when we have time to refactor the widgets.

Bluesmile82:
The app is crashing when:

- When on a profile (I tried a place profile) go to change profile
- Click on importer country

I'm not sure if this is related to the PR. Its actually because at some point we dont have a context and therefore we cant find the name on profile.component

image

frontend/scripts/react-components/profile/profile-widgets/top-consumers-widget.component.jsx
Bluesmile82: Actually the idea was to slowly change the classes to functions and hooks whenever we edit the code. But it's ok

tomasmoose:
> The app is crashing when:
> 
> * When on a profile (I tried a place profile) go to change profile
> * Click on importer country
> 
> I'm not sure if this is related to the PR. Its actually because at some point we dont have a context and therefore we cant find the name on profile.component
> 
> image

It was not directly related, but as we no longer require "context" some of our prop checks failed + some widgets we need to double check we have a context, otherwise they will fail.

Bluesmile82:
Probably we have to check all the cases for the lack of context. I will prefer to add a new task for this. But the purpose of this PR is reached so Ill merge it and deploy

tomasmoose:
> Probably we have to check all the cases for the lack of context. I will prefer to add a new task for this. But the purpose of this PR is reached so Ill merge it and deploy

Yes i think we need to check it a bit more deeply, maybe we need a small refactor on the profile component

PR #1530, opened 2020-09-04
Schema rename

harrybiddle:
## Description

Renames `tool_tables` to `trase_earth` to reflect the recent name change on Trase's systems.

@agnessa I presume we will need to update the environment variables on the various servers too? All future dumps we make will be using `trase_earth`, so we are good to do this 👍


agnessa:
Nice, thanks! The env var is updated already in staging, sandbox, demo and indonesiademo - we don't run the database update in production, so not needed there. Can you rebase the branch with develop (I think maybe you branched off master and it picked some merge commits?) and squish the two commits with the rename?

harrybiddle:
Right you are! I was off master. Rebased and squished 👍

PR #1567, opened 2020-10-02
fix crash when levels are not available on explore step 3

tomasmoose:
## Pivotal Tracker

https://www.pivotaltracker.com/n/projects/2380432/stories/175073497

## Description

This pr fixes crash happening on step 3 explore when levels are not available. 

## Testing instructions

Api URL needs to point to sandbox to reproduce, make sure step 3 does not crash.


Bluesmile82 (approved):
Thanks!

PR #1578, opened 2020-10-13
Reset dimensions on change context to enable default

Bluesmile82:
## Description

image

We were setting the default for the dimensions (seen on the choropleth) when the dimensions were empty but this only happened the first time we accessed the tool. After changing the context we should have the same effect so we reset the dimensions when we finish selection on the tool panel.

We still keep the blank selection if the user deselects all dimensions. 

## Testing instructions

Go to a context then select a different one from the tool node panel. The default choropleth should show.


tomasmoose (commented):
✅ Looking good

PR #1580, opened 2020-10-13
Fix/unit layers data environments

Bluesmile82:
## Description

This PR adds a new env variable for the request of the Unit layer data endpoint. This data is different in each environment, therefore the env variable. Also, we want to be able to test different environments on development

image

## Testing instructions

Go to Indonesia - Palm Oil. On Kabupaten column and with one unit layer indicator selected you should be able to see data on the map tooltip. There wasn't any before.


tomasmoose (commented):
✅ Looking good

PR #1581, opened 2020-10-13
Fix/valid recent context

Bluesmile82:
## Pivotal Tracker

https://www.pivotaltracker.com/story/show/175242862

## Description

Fix cases in which the latest context was not included on the currently available contexts

### DIFF
--- a/frontend/scripts/react-components/explore/explore.selectors.js
+++ b/frontend/scripts/react-components/explore/explore.selectors.js
@@ -130,7 +130,6 @@ export const getCardsWithRecentCard = createSelector(
   [getSankeyCards, getRecentCard, getContexts],
   (cards, recentCard, contexts) => {
     const { countryId: recentCountryId, commodityId: recentCommodityId } = recentCard.data[0];
-    console.log('c,', recentCountryId, recentCommodityId)
     const recentValidContext = contexts.find(context => recentCountryId === context.countryId && recentCommodityId === context.commodityId);
     if (!recentValidContext) return cards;
     if (!cards) return recentCard;
--- a/frontend/scripts/react-components/explore/explore.selectors.js
+++ b/frontend/scripts/react-components/explore/explore.selectors.js
@@ -107,7 +107,7 @@ export const getCountryQuickFacts = createSelector([getQuickFacts], quickFacts =
   return countryQuickFacts;
 });
 
-const getRecentCard = () => {
+export const getRecentCard = () => {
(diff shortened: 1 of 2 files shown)


frontend/scripts/react-components/explore/explore.selectors.js
```diff
@@ -122,13 +122,17 @@ const getRecentCard = () => {
...
 }
 
-const getCardsWithRecentCard = createSelector(
-  [getSankeyCards, getRecentCard],
-  (cards, recentCard) => {
+export const getCardsWithRecentCard = createSelector(
+  [getSankeyCards, getRecentCard, getContexts],
+  (cards, recentCard, contexts) => {
+    const { countryId: recentCountryId, commodityId: recentCommodityId } = recentCard.data[0];
+    console.log('c,', recentCountryId, recentCommodityId)
```
tomasmoose, line 133:
```suggestion
```

frontend/scripts/tests/explore/explore-selectors.test.js
```diff
@@ -0,0 +1,43 @@
...
+
+    const selector = getCardsWithRecentCard;
+    beforeEach(() => {
+      selector.resetAllComputations;
+      jest.resetAllMocks();
+    });
+
+    const setMock = (countryId, commodityId) => jest.fn(() => `${countryId}-${commodityId}`);
+    it('returns recent card if it has a valid context', () => {
+        Storage.prototype.getItem = setMock(validCountryId, validCommodityId);
```
tomasmoose, line 32: Recommend adding this in the global jest config, so we don't have to care about local storage, 

```
class LocalStorageMock {
  constructor() {
    this.store = {};
  }

  clear() {
    this.store = {};
  }

  getItem(key) {
    return this.store[key] || null;
  }

  setItem(key, value) {
    this.store[key] = value.toString();
  }

  removeItem(key) {
    delete this.store[key];
  }
};

global.localStorage = new LocalStorageMock;
``` 

This is basically all you need to implement the necessary requirements for the API

Bluesmile82:
Hi! Thanks for the comments! I have added the global localStorage mocking. The second test is still not working. We might want to find out why but I have to merge as they are experiencing this bug and they need to test sandbox

tomasmoose:
> Hi! Thanks for the comments! I have added the global localStorage mocking. The second test is still not working. We might want to find out why but I have to merge as they are experiencing this bug and they need to test sandbox

no worries, I would rather remove the test as now we probably wont fix it as its merged

PR #1607, opened 2020-11-04
Feature/full screen button

Bluesmile82:
## Description

This PR adds a full-screen button only to the iframe contained application.
This button once pressed uses the Full-Screen API to expand the app to full size.

It's mandatory that the iframe has the allowfullscreen and allow="fullscreen" attributes

Caveat: The browser is expanded to full-screen mode. I don't think this was part of the requirement but I don't think there its another way with the Full-screen API.

## Testing instructions

- Run a server in the tests/iframe folder. 
- Go to the port that the server serves (in my case 8000). 
- Press the full-screen button: It should expand and change to the exit button. 
- It should exit the full screen when the exit button or ESC key is pressed.

full


tomasmoose (approved):

tomasmoose (approved):
Works great; 

```
It's mandatory that the iframe has the allowfullscreen and allow="fullscreen" attributes
```

Just make sure they know about this! 


```
Caveat: The browser is expanded to full-screen mode. I don't think this was part of the requirement but I don't think there its another way with the Full-screen API.
```

This is the way, so don't worry, if they want another solution its out of our hands

tomasmoose (approved):

PR #1625, opened 2020-11-18
Fix/subnational commodities

Bluesmile82:
## Description

Some commodities weren't showing as subnational because only the first context of that commodity was checked for subnational years

image

## Testing instructions

Go to explore page. Commodities like Wood Pulp should be red


tomasmoose (approved):
Great find!

PR #1631, opened 2020-11-24
Improve pluralize

Bluesmile82:
https://app.asana.com/0/1187619191620560/1199244035527549/f

This PR improves the pluralize function using the pluralize package and pluralizing by default the last word of a string. The exceptions are less now.

image

## Testing instructions

Go to the data section of Ecuador - Shrimp. You should see all correct plurals, including 'parishes' and 'exporter groups'


tomasmoose (approved):
nice solution 👍

PR #1650, opened 2020-12-03
Alter the node type that indonesia concession boundaries are associated with

harrybiddle:
## Pivotal Tracker

Please include the link(s)

## Description

Previously Indonesian "concession boundaries" (i.e. large wood plantations) were linked to the node type `CONCESSION`. After a request from our external partners these boundaries should now be linked to the node type `WOOD_SUPPLIER`.

As far as I can tell this is how you can achieve that!

**If this PR is OK and gets merged it would be great if it could be deployed to sandbox.**

## Testing instructions

To test this requires the right data: you would need to have the data from https://sandbox.trase.earth, and then to go Indonesia > Wood Pulp and confirm that the column `WOOD_SUPPLIER` in the Sankey is linked to the concession boundaries in the map.


Bluesmile82 (approved):
Thanks! This file is actually just a backup for the layer that has to be uploaded to the server. I just uploaded it and it's working correctly 👍

PR #1693, opened 2021-01-14
Update footer design

Bluesmile82:
## Designs

https://projects.invisionapp.com/d/main/#/console/20810309/437845102/inspect

image

## Description

image

## Testing instructions

The form submit won't work as we don't send the last name and we send country instead. They should update their form. Maybe we should add the last name too

### DIFF
--- a/frontend/scripts/react-components/shared/newsletter/newsletter-form.component.jsx
+++ b/frontend/scripts/react-components/shared/newsletter/newsletter-form.component.jsx
@@ -9,6 +9,7 @@ import cx from 'classnames';
 import Heading from 'react-components/shared/heading/heading.component';
 
 import 'react-components/shared/newsletter/newsletter.scss';
+import { startCase } from 'lodash';
 
 class NewsletterForm extends React.PureComponent {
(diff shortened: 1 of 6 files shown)


frontend/scripts/react-components/shared/newsletter/newsletter-form.component.jsx
```diff
@@ -19,7 +20,8 @@ class NewsletterForm extends React.PureComponent {
         email: '',
         firstname: '',
         lastname: '',
-        organisation: ''
+        organisation: '',
+        country: ''
```
tomasmoose, line 24: @Bluesmile82 I think you need to add "country" here as well https://github.com/Vizzuality/trase/blob/develop/app/controllers/api/v3/newsletter_subscriptions_controller.rb#L13 

Then we should notify them about these changes just to make sure they catch them

Bluesmile82:
I updated the fields to be able to accept both forms. Take a look if you can @tomasmoose

tomasmoose (approved):
Looks good! merging

PR #1699, opened 2021-01-19
uppercase company titles

tomasmoose:
https://app.asana.com/0/1195054825337974/1199601330969645/f

### DIFF
--- a/app/services/api/v3/actors/basic_attributes.rb
+++ b/app/services/api/v3/actors/basic_attributes.rb
@@ -180,7 +180,7 @@ module Api
 
         def summary_of_total_trade_volume(profile_type)
           if @trade_total_current_year_raw.zero?
-            return "<span class=\"notranslate\">#{@node.name.humanize}</span> \
+            return "<span class=\"notranslate\">#{@node.name.upcase}</span> \
             #{profile_type.first(-1)}d 0 tons of #{@commodity_name} from \
             #{@context.country.name} in \
 <span class=\"notranslate\">#{@year}</span>."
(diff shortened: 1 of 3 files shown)


app/services/api/v3/actors/basic_attributes.rb
```diff
@@ -180,7 +180,7 @@ def importer_summary
 
         def summary_of_total_trade_volume(profile_type)
           if @trade_total_current_year_raw.zero?
-            return "<span class=\"notranslate\">#{@node.name.humanize}</span> \
+            return "<span class=\"notranslate\">#{@node.name.upcase}</span> \
```
agnessa, line 183: I think there are 3 more instances of `@node.name.humanize` in this file that should be changed in the same way. Quickest way to test this in development is to delete the computed summaries in psql:
`rails dbconsole`
`> UPDATE nodes_with_flows SET actor_basic_attributes = NULL;`
Longer version would be to recompute the summaries in development, rather than delete them:
`rails console`
`> Api::V3::RefreshProfiles.new.call`

In sandbox / production best to use the background jobs version:
`RAILS_ENV=sandbox bundle exec rake basic_attributes:refresh_later`

tomasmoose: done, I think I made it work locally so we would run this when merged right? 

```
RAILS_ENV=sandbox bundle exec rake basic_attributes:refresh_later
```

agnessa: after it's deployed, that needs to be run on `sandbox-data`

agnessa (approved):

Bluesmile82 (approved):

PR #1762, opened 2021-03-02
Fix continuous deployment

agnessa:
Fixed locked Capistrano version, which was breaking continuous deployment.

Also, added a dependency on the front_end job when deploying.

Screenshot 2021-03-03 at 09 59 07


tomasmoose (approved):
Brilliant will merge! don't worry about the lining warnings, this is a good thing, we will address them.

PR #1787, opened 2021-04-09
Calculated HDI ranking

agnessa:
## Asana

https://app.asana.com/0/0/1200124803845199/f

## Description

Tries to calculate the rank relative to other countries. So far ends up with incorrect results due to missing data.


agnessa:
Obsolete, we display the score since placement cannot be reliably calculated.

